PR #11, opened 2015-09-17
enable_extension "uuid-ossp" needs to be run at least once on the dev/prod database

TheNotary:


TheNotary:
validating emails.. there's some content addition of "uh something went wrong", etc, someone might want to tweak to setup with the proper voice.

Hainish:
@TheNotary looks like this includes a lot more than just the uuid-ossp change.  Can you isolate this change in a new PR?

PR #18, opened 2015-09-23
fixed bug where when user creates a partition will all empty fields, …

TheNotary:
…it's invalid thus is nil, yet the analytics page tries to create a csv path out of it, crashing view


TheNotary:
We were drunk and merged this.  It's OK!

PR #24, opened 2015-10-08
Making csv more useful

TheNotary:


app/controllers/application_controller.rb
```diff
@@ -25,6 +25,22 @@ def manifest
...
+
+    # We now have one or two parts (depending on whether we could find
+    # a suitable period). For each of these parts, replace any unwanted
+    # sequence of characters with an underscore
+    fn.map! { |s| s.gsub /[^a-z0-9\-]+/i, '_' }
+
+    # Finally, join the parts with a period and return the result
+    return fn.join '.'
+  end
+
```
Hainish: For generic functions such as this, instead of putting it in a controller it makes more sense to put it in a helper.

app/models/signature.rb
```diff
@@ -39,6 +39,73 @@ def time_ago
...
+      "Texas" => "TX",
+      "Utah" => "UT",
+      "Vermont" => "VT",
+      "Virginia" => "VA",
+      "Washington" => "WA",
+      "West Virginia" => "WV",
+      "Wisconsin" => "WI",
+      "Wyoming" => "WY"
+    }
+  end
```
Hainish: seems like this belongs in its own model, since it can be used generically and doesn't have to do with singatures specifically

PR #35, opened 2015-10-29
limit user to 20 login attempts before 2hr timeout

TheNotary:


TheNotary:
Fastly might mess up our flash message.

PR #43, opened 2015-12-03
Cucumber testing

TheNotary:


TheNotary:
I can squash these 'itemized' commits down at request.  They might illustrate the story of implementing as is though.

PR #46, opened 2015-12-07
hiding twitter character count because it was wrong

TheNotary:
This commit is running on production and seems fine.  It's tricky because the admin page uses this function, and so does the user page.  It's non-ideal for on the user page because the 'estimate' of characters left is wrong thus https://redmine.eff.org/issues/8590.

### DIFF
--- a/app/assets/javascripts/action_page.js
+++ b/app/assets/javascripts/action_page.js
@@ -44,11 +44,15 @@ $('input:radio').screwDefaultButtons({
   height: 15
 });
 
+// Give an approximate character count guide to see how many character's are
+// left in a tweet or social media card.  Tweet length isn't so accurate due to
+// url shorteners and handle names (if they're inserted after the tweet is
+// crafted)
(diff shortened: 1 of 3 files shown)


mfb:
Are you familiar with the Twitter text libraries?

We use it on kittens.eff.org to count tweet characters

https://github.com/twitter/twitter-text

TheNotary:
Though unfamiliar, I was considering it, but based on Rainey's post and chats with max about how unliked the feature is, I didn't want to make a step towards increasing the amount of javascript included in the library...  But the ticket didn't discuss the calculation that's done on the admin page which I kept in place, so now I'm uncertain whether to leave that in as the janky version, strip it out completely or consume twitter's twitter-text library  :(

TheNotary:
Code for stubbing social media icon callback in test env had a typo ><

TheNotary:
Needed to check that a flash message got through fastly correctly.  I think those are in the session cookie, so should be fine...

Issue #47, opened 2015-12-08
Staging Server tests of commits after 57e7286


TheNotary:
Other than the two javascript/ css cherry picks, the last stable commit pushed to production was...

    57e7286 Building in support for new recaptcha-style forms

I'm going to push all these commits to the staging server, diagnose them carefully, listing any problems in this thread, and then upon resolution, will push them to the production server.  

Commands to undo the commit incase it's fubar:
```
#  On production SSH
$  rake db:rollback STEP=1

# On dev git repo...
$  git push production 9e9c784:master
```

That should put things right in less than a minute or so of downtime.

PR #48, opened 2015-12-08
fixed bug revealed via the staging server

TheNotary:


TheNotary:
This resolves the staging checks #47

Issue #84, opened 2016-05-23
Conditionally show all signatures on action pages


vbrown608:
When "show all signatures on the action page" is checked:
- signatures should appears in the main column below the petition
- signatures should be paginated
- the "recent signatures" teaser in the right column should not be visible

PR #103, opened 2016-06-01
Show all signatures

squiggy:
Related to #84


vbrown608:
Looks good! Could you add a couple tests? I think one for "show all" enabled and one for "show all" disabled would be good.

Other than that, just a few tiny nits:
- typo in line 2 of petition.rake
- in line 32 of action_page/show.petition.erb, `@actionPage.petition` should probably go before `@actionPage.petition.show_all_signatures`
- it would be cool to match the pagination to the mockup. not sure how much flexibility rails built-in pagination allows, but it's worth a shot. that can probably be its own ticket and pull request, just don't want to forget.

PR #113, opened 2016-06-16
embedded action parameter for email tool non-US option

mfb:
Add a parameter for embedded actions to provide a button to hide the email tool and display the petition tool

### DIFF
--- a/app/assets/javascripts/application/tools/email.js.erb
+++ b/app/assets/javascripts/application/tools/email.js.erb
@@ -28,6 +28,12 @@ $(document).on('ready', function() {
     });
   };
 
+  $('#non-us-option').click(function () {
+    $('#petition-tool').show();
+    $('#email-tool').hide();
+    $('#signatures').show();
+  });
+
   var generateCongressForm = function (options) {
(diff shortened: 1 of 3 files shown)


vbrown608:
Tests pass locally. Under the current setup, they won't pass in Travis until the Dockerize branch is merged :(

I think it would be railsier to copy to params[:non_us_option] to an instance variable in the controller and use that instance variable in the view. See http://stackoverflow.com/questions/15823695/best-practice-passing-instance-variables-or-using-params-in-ruby-on-rails-view. In this case, I believe the controller action is action_page#embed_iframe - see handling of @css there.

If you agree it would be cool to push a tweak :)

In the meantime, I'm going to merge this so the latest campaign can launch :+1:

mfb:
Agree this is messy/hacky however I was copying how other variables are being used. So we should probably fix all of them at the same time.

Side question: is it Railsy to have javascript built by a view - e.g. views/action_page/embed.js.erb? This feels like spaghetti code to me - I like to have static pure javascript files and make data available to the javascript separately.

PR #115, opened 2016-06-18
Build out institutions + affiliations

vbrown608:
- Create the affiliation model + add it to action_pages via has_many
- Add institutions and affiliations to signatures
- Prevent duplicate institutions

A petition action with institutions, affiliations, and signatures can now be created with rake petition:create_local.


squiggy:
Looks good!

Issue #118, opened 2016-06-20
Add popover notice to the end of petition and email actions


maximillianh:
Please add the following text to the end of petition and email actions:

> How will EFF use this information? (?)

When the user mouses over the question mark, they should see:

> We use this information to help you contact decision makers (e.g. Members of Congress) and their staff. We may work with other groups to assist in this action, such as sharing and jointly delivering petition signatures with a coalition partner. We may use Action Center information to assess the success of action campaigns and to improve the functionality and effectiveness of our site, and to allow you to see records of your activities. We, and our coalition partners, may publicize and share aggregate information about action campaigns, such as the number of people who participated in a particular location or region or through a type of action, as well as sharing or publicizing activity trends.

mfb:
Just now pushed the CSS for this to the branch.

PR #121, opened 2016-06-21
apply CSS for privacy notice refs #118

mfb:


mfb:
Someone should take a look at the styling here

squiggy:
It needs more space above the blue text. It is covering the checkbox for "Don't add my name or city to this petition (we'll only list your country)." 

I pushed an edit for this in feature/privacy-notice_squiggy

squiggy:
I cleaned up my commit and force pushed to feature/privacy-notice_squiggy

mfb:
@squiggy you added 6px top margin to all the sections of the email tool.   Personally I think it looks better the way it was before (top margin 0).  But I don't want to bike shed...

squiggy:
That makes sense. Actually the issue I noticed is pretty minor. The text overlaps the checkbox slightly when logged in while using Firefox. It looks fine in Chrome. 
screenshot

mfb:
ok yeah that looks bad enough to fix, I would say.

PR #167, opened 2016-08-27
Readme overhaul

vbrown608:
Notable changes:
- Setup instructions use Docker now
- Make language more platform-y and less EFF-specific
- Change details that aren't true anymore (ex maintainers)
- More consistent whitespace


vbrown608:
References #83

Issue #168, opened 2016-09-02
Clarify user management docs


vbrown608:
It wasn't clear to techops whether the [ ] should be included in `rake users:add_admin[name@example.com]`, or whether they denote text that should be replaced. We should add another example to the readme.

vbrown608:
Also, after running the command users see
```
** [Raven] Raven 0.15.6 configured not to send errors.
** [Raven] Raven 0.15.6 ready to catch errors
```

Which isn't very encouraging.

PR #172, opened 2016-09-09
Improve setup workflow

vbrown608:
Based on feedback from Rocket and Peter
- Separate docker-compose build and docker-compose up steps
- Clarify user adminification docs and make success successier (#168)
- Don't recommend docker-compose auto-restart in dev environments


wioux:
+1

Issue #176, opened 2016-09-12
RestClient requests


wioux:
Wrap calls to external services in a PORO or module. This is already done in the `SmartyStreets` module, but should be done in other areas too e.g. [this part of ToolsController#call](https://github.com/EFForg/action-center-platform/blob/43eb4528b308755f4499b0f2506b69d44bd4a872/app/controllers/tools_controller.rb#L30-L57)

```ruby
    begin
      response = RestClient.get Rails.application.config.call_tool_url + '/call/create',
        params: { campaignId: params[:call_campaign_id],
                  userPhone:  params[:phone],
                  userCountry: 'US',
                  userLocation: params[:location],
                  # TODO - Settle on the schema of the private meta data
                  meta: {
                    user_id:     @user.try(:id),
                    action_id:   params[:action_id],
                    action_type: 'call'
                  }.to_json,
                  callback_url: root_url }
    rescue RestClient::BadRequest => e
      begin
        error = JSON.parse(e.http_body)["error"]
      rescue
        raise e
      end
      # Don't raise for twilio error 13224: number invalid
      unless error.match(/^13224:/)
        if Rails.application.secrets.sentry_dsn.nil?
          raise error
        else
          Raven.capture_message(error, { :level => 'info' })
        end
      end
    end
```

We can also consider caching responses from SmartyStreets and Sunlight when they only depend on zip code.

PR #183, opened 2016-09-13
Create CallTool module to handle rest interactions with call tool

wioux:
This branch encapsulates interactions with the call tool api in a new module named `CallTool` (addressing #176). Please note that it's untested since I'm not able to hit the api endpoint in development.


vbrown608:
Nice!

Issue #185, opened 2016-09-14
Find a happy home for rake assets:precompile


vbrown608:
It used to be added to "command" in the production Dockerfile, ex `command: bash -c unicorn && rake assets:precompile`

That caused about 15 seconds of downtime on deploy as the assets compiled.

I moved it to the Dockerfile a few weeks, since it seems to fit with building the application image. That started causing problems because .js.erb files relied on environment variables that were not set at the time the container was being built.

I moved it (and `rake webshims:update_public`) to docker/entrypoint.sh for now, but that will still cause downtime when we deploy. Need to find a better solution.

vbrown608:
@wioux and I discussed two approaches to solving this problem:

1. Remove references to environment variables from all files with /assets and precompile assets on build. Two drawbacks to this solution:
  - Future contributors might not know this will create problems
  - It mixes the dev and production setups by always requiring assets to be precompiled

2. Precompile assets in entrypoint.sh and use the load balancer to avoid downtime. Example described here: https://github.com/docker/compose/issues/734#issuecomment-167392770

We'd like to discuss with @actualben when he gets back.

PR #186, opened 2016-09-14
Precompile assets in entrypoint.sh

vbrown608:
See #185 for details


vbrown608:
@wioux let's hold off on merging this and see if we can find a better solution tomorrow

PR #187, opened 2016-09-19
Pass image_gallery.feature by testing against amazon bucket from config

wioux:
This fix lets cucumber tests pass when using a more generic `docker-compose.yml`.


vbrown608:
Yay improving tests

PR #190, opened 2016-09-20
Javascript cleanup

wioux:
This branch moves the inline javascript present in some views into corresponding scripts under app/assets.


vbrown608:
Oops, still getting used to Github's new pull request review tools. That was a comment on 79a966f in particular.

PR #201, opened 2016-09-22
Refactor action page controller set_action_display_variables method

vbrown608:
Start refactoring action page controller set_action_display_variables method by pulling redirects into separate before_filters. Feedback welcome!


wioux:
Can we rename `redirect_target_action_if_archived` to something without the `_if_archived` suffix?. To me it looks odd since this is the usage:

```ruby
  if @action_page.redirect_from_archived_to_active_action?
    redirect_to @action_page.redirect_target_action_if_archived
  end
```

`action_page_redirect_target`? I don't know. Overall I like this direction though.

Issue #188, opened 2016-09-19
enable admins to delete petition signatures


mfb:
petition signatures are subject to spam/vandalism/etc. so admins should be able to delete signatures. 

The petition list page, e.g. https://act.eff.org/admin/petitions/85 could have a filter text box, checkboxes, and a check-all checkbox, for bulk deletion of signatures, as well as delete links for individual signatures (needless to say with CSRF protection... :)

PR #204, opened 2016-09-23
Petition signature deletion

wioux:
This branch adds a search feature to the page on `/admin/petitions/:id` so that signatures with a particular name or email can be listed. The search is case insensitive and allows for partial matches. It also provides a button on that page for deleting signatures, fixing #188.

### DIFF
--- a/app/assets/javascripts/admin.js
+++ b/app/assets/javascripts/admin.js
@@ -107,6 +107,10 @@ $(document).on('ready', function() {
     history.pushState(null,null,'#' + anchor);
   });
 
+  $(".container input[type=checkbox].select-all").click(function(e) {
+    $(this).parents("form").find("input[type=checkbox]").prop("checked", this.checked);
+  });
+
   var updated_at = $('#action_page_updated_at');
   if (updated_at.length > 0) {
     var t = window.setInterval(function() {
--- a/app/assets/javascripts/admin.js
+++ b/app/assets/javascripts/admin.js
@@ -8,6 +8,7 @@
 //= require admin/file_uploads
 //= require admin/s3_uploads.js
 //= require admin/email_tabulation
+//= require admin/petitions
 //= require s3_cors_fileupload
 //= require bootstrap-sass/bootstrap/tab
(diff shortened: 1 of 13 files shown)


vbrown608:
This looks good! Thanks for removing packaged dependencies from the dockerfile :)

Any idea about the source of this strange new cucumber output from features/users/create_content.feature:11?
```
t
th
thi
this
this 
this i
this is
this is 
this is a
this is an
this is an 
this is an i
this is an im
this is an imp
this is an impo
this is an impor
this is an import
this is an importa
this is an importan
this is an important
this is an important 
this is an important c
this is an important ca
this is an important cal
this is an important call
```

vbrown608:
Filtering works great! 

I recommend adding a few css classes for visual consistency with the rest of the site:
- btn-danger (rather than btn-cancel and btn-default) class on the "Destroy" button to match other delete buttons on the site
- btn, btn-sm, and btn-default classes on the "Search" button
- form-control and input-sm class on the search field

The function of the "select all" box wasn't obvious to me. We'll also need to make it clear to users whether they're selecting all the signatures for the petition, just the ones that match the current filter, or just the ones that match the filter and are visible on the current page. Happy to do some co-planning on that if you want.

Last thing - I suspect that the check boxes will need either a label or a column header to make them accessible to users with screen readers. I think a visually hidden label element that says something like "select for deletion" would work well. I'm not sure if bootstrap provides a good class for that, the action center already has one, or we'd need to create our own (something like http://www.zeldman.com/2012/03/01/replacing-the-9999px-hack-new-image-replacement/)

vbrown608:
I did some more reading about accessibility for checkboxes in tables last night. Made a tiny tweak to match the 508 docs: http://www.section508.va.gov/support/webdev/2_7a.asp

I also realized a visually-hidden label might be helpful for the search bar. Bootstrap 3 provides the sr-only (screen reader only) class for that.

I think this is ready for merge now!

wioux:
:+1: including the signature name in the select title is a nice touch.

Issue #194, opened 2016-09-20
tools_controller email_target is using incorrect method for URL-encoding


mfb:
apparently URI::escape is old and broken. it doesn't URL-encode commas, for example.  

This causes mailto: redirect using the default mail client to fail on Chrome browser.

It looks like we should be using http://ruby-doc.org/stdlib-1.9.3/libdoc/erb/rdoc/ERB/Util.html#method-c-url_encode for URL-encoding.

mfb:
forgot to mention this is being tracked internally in https://redmine.eff.org/issues/9988 so when fixed we should close it there too

wioux:
Fixed in #205.

PR #205, opened 2016-09-26
Kill CGI::escape

wioux:
This branch replaces escaping done by `CGI::escape` with `ERB::Utils#url_encode` (which is aliased as `u`), fixing #194. I created a `service_uri` method for email campaigns, which responds with a mailto or compose url for the given mail provider, and I've also updated the format for outlook urls, since the current links seem to no longer work.

### DIFF
--- a/app/controllers/tools_controller.rb
+++ b/app/controllers/tools_controller.rb
@@ -162,25 +162,11 @@ class ToolsController < ApplicationController
         { type: "action", actionType: "email", actionPageId: params[:action_id] },
         action_page: @action_page
     end
-    subject_uri = URI::escape @action_page.email_campaign.subject
-    subject_cgi = CGI::escape @action_page.email_campaign.subject
-    body_uri = URI::escape @action_page.email_campaign.message
-    body_cgi = CGI::escape @action_page.email_campaign.message
-    body_hotmail = CGI::escape @action_page.email_campaign.message.gsub("\r\n","<br>")
-    to_cgi = CGI::escape @action_page.email_campaign.email_addresses.gsub(" ","")
-    case params[:service]
-    when "default"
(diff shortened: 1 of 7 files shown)


mfb:
mailto: URIs need to use RFC 3986 encoding: spaces encoded as %20 rather than +

mfb:
I hate to be pedantic but :) technically speaking the email addresses which are query parameters should be URL-encoded. 

(Whereas the email addresses in the mailto: URLs are serving their reserved character purpose so they are fine as-is :)

wioux:
Thanks @mfb. :+1: robustness

mfb:
ok this looks good to me, but at the moment I'm not set up to test it on an actual action page, e.g. https://act.eff.org/action/stop-the-government-s-tattoo-recognition-experiments

wioux:
Just tested against that page. mailto/gmail work fine, but outlook is broken. It seems they `400 Bad Request` if the URL is longer than 2100 chars. If I go back to `+` encoding spaces in non-mailto URI's then the compose URL length for that page will be small enough to work again, but in general this is not a problem we can fix.

mfb:
well then maybe a good idea to convert %20 to + for non-mailto URLs...  what a PITA

found one other nit.  http://compose.mail.yahoo.com/ should be https://compose.mail.yahoo.com/ :)

wioux:
Alrighty outlook works again. I didn't give the yahoo links any love since they have been [disabled](https://github.com/EFForg/action-center-platform/blob/master/app/views/tools/_email.html.erb#L31-L39) since the initial commit. If we want to bring them back I'll check it out.

mfb:
I guess yahoo is disabled because line breaks in the email body didn't work (according to comment)

mfb:
if email action works then let's ship it (if this code is also used for twitter actions then I guess QA that too)

vbrown608:
Looks good to me!

PR #207, opened 2016-09-27
Fix call tool for custom campaigns

wioux:
Call actions currently break when the campaign is configured as "custom" on call.eff.org. In that case the tool doesn't require any location data, and as a result action center doesn't append any inputs to the call form (even for phone number)

[call.js](https://github.com/EFForg/action-center-platform/blob/master/app/assets/javascripts/application/tools/call.js#L27-L35)
```javascript
        if(required_location == "postal"){
          $container.append($phone_number_field);
          $container.append($zip_field);
        }
        if(required_location == "latlon"){
          $container.append($phone_number_field);
          $container.append($street_address_field);
          $container.append($zip_field);
        }
        // ... no catch for required_location == undefined
```

This branch fixes the bug. I no longer ask for the campaign's required fields in an ajax request, and instead just create the corresponding form elements on the initial page serve. When the call campaign is custom, only a phone number input is required.

### DIFF
--- a/app/assets/javascripts/application/tools/call.js
+++ b/app/assets/javascripts/application/tools/call.js
@@ -3,38 +3,13 @@ $(document).on('ready', function() {
     height_changed();
 
     var call_campaign_id = $('[data-call-campaign-id]').attr('data-call-campaign-id');
-    var url = '/tools/call_required_fields?call_campaign_id=' + call_campaign_id;
-    var required_location;
-
-    var $phone_number_field = $(document.getElementById('phone-input').text);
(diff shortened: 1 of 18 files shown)


vbrown608:
When I visit `http://localhost:3000/action/sample-call-action` I see "uninitialized constant ActionView::CompiledTemplates::CallTool"

wioux:
Whoops. I've fixed this by adding `lib/` to the autoload search path.

vbrown608:
Nice solution!

PR #209, opened 2016-09-29
Create rake task for deleting petition signatures with non-unique email

wioux:
This branch implements the idea suggested in #3 -- that there should be a rake task for deleting duplicate signatures.


vbrown608:
Hooray for our inaugural test of a rake task!

PR #210, opened 2016-09-29
Legislator lookup

### LINKED ISSUE #203: admin: javascript error due to legislators missing from congress-zip-plus-four legislators.json
https://rawgit.com/EFForg/congress-zip-plus-four/master/legislators.json is used on the admin pages of act.eff.org to generate the analysis tables, e.g. https://act.eff.org/admin/email_campaigns/177/congress_tabulation

However it is missing some members of congress, e.g. K000388 D000625 L000585  which results in a javascript fatal error at some tabulations, e.g. https://act.eff.org/admin/email_campaigns/166/congress_tabulation

1.  Regenerate legislators.json (not sure how this is done but hope it's in the README). Side note:  We need to make sure that legislators.json contains all historical legislators, not just current legislators, otherwise we'd get fatal errors in the future.
1. Fix this in JS so if a legislator is missing, there is no fatal error and results can still be rendered.

wioux:
This branch fixes #203.

For action center purposes we don't need all of the fields out of [legislators.json](https://github.com/EFForg/congress-zip-plus-four/blob/master/legislators.json), only a map from bioguide id to legislator name. The rest of the fields (street addresses of each congressional district office) are comparatively difficult to locate/maintain, so I propose using an alternative data source for looking up legislator names in action center.

The project at [unitedstates/congress-legislators](https://github.com/unitedstates/congress-legislators) is actively maintained and contains the necessary information. This branch adds a `Congress` module which fetches data from that project and culls/formats it for use in the email/staffer tabulation actions.


mfb:
sweet, I was thinking the same thing...

vbrown608:
Looks good to me! :eyes:

PR #212, opened 2016-09-30
Set Vary: Accept-Encoding, Accept-Language, Cookie

wioux:
Set Vary HTTP header and include Accept-Language.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -66,6 +66,6 @@ class ApplicationController < ActionController::Base
   end
 
   def set_cache_headers
-    response.headers["Vary"] = "Accept-Encoding, Accept-Language, Cookie"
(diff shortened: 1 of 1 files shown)


mfb:
hrm this sounds bad for cacheability. do we actually vary language based on the Accept-Language header?  I'd prefer to change the URL when serving content in a different language.

wioux:
We [set locale](https://github.com/EFForg/action-center-platform/blob/master/app/controllers/application_controller.rb#L64) based on the Accept-Language header. This creates a bug at the moment where if you're the "first" visitor to a page, everyone thereafter sees the page in the locale that was rendered for you (until the cached page expires).

mfb:
lol yes that would.

ok well maybe we can revisit this in future: would be nice to improve cacheability and allow people to share URLs in specific languages by altering the URL per-language.

vbrown608:
:cookie:

PR #213, opened 2016-09-30
Newsletter signup on actions

### LINKED ISSUE #110: Add newsletter signup for all actions
I think it would be useful to optionally be able to add an email signup for actions that don't require an email address. Specifically:

- Twitter
- Calls
- Emails that use the user's own email client (i.e., not emails to Congress)

wioux:
This is a branch I've been working on to fix #110 and #1. It adds additional newsletter opt-ins to email, call, and tweet actions. On email and call actions these are styled the same as the opt-in on petitions, and on tweet actions it's a separate form at the bottom of the action which submits by ajax.

email example com eff action center

call-action

tweet at eff eff action center

tweet at eff eff action center 1


app/assets/stylesheets/action_page.css.scss
```diff
@@ -434,6 +433,10 @@ h3.privacy-notice-header {
   }
 }
 
+.email-signup {
+  text-align: left;
```
vbrown608: Can also use the bootstrap class "text-left" https://getbootstrap.com/css/#type-alignment

app/assets/javascripts/application/tools/newsletter_signup.js
```diff
@@ -0,0 +1,30 @@
+$(document).on("ready", function() {
+  $("form.newsletter-subscription").on("ajax:beforeSend", function() {
+  });
+
+  $("form.newsletter-subscription").on("ajax:complete", function(xhr, data, status) {
+    if (status == "success") {
+      $(this).find(".radio-inline, input").attr("disabled", true);
+      $(this).find("input, .radio-inline").css("opacity", 0.5);
```
vbrown608: I think it would be better to have the opacity and background color in the css, based on the presence of the "disabled" class.

app/assets/stylesheets/action_page.css.scss
```diff
@@ -377,8 +374,10 @@
 .description h3, h3.privacy-notice-header,
 h3.header, #tools #email-tool h3 {
   position: relative;
-  margin-bottom: 1em;
+  margin-bottom: 1em !important;
```
vbrown608: Instead of using !important here, perhaps you could make a change to line 589 action_page.css.scss to increase (or not define) the margin there.

app/assets/stylesheets/action_page.css.scss
```diff
@@ -266,7 +263,7 @@
   text-align: center;
 
   .privacy-notice {
-    margin-bottom: 0;
+    margin: 10px -10px 0 -10px;
```
vbrown608: I couldn't find where this is being rendered. If it's needed and it's possible to do it without negative margins, that would be ideal. I've found that they sometimes cause confusion down the line. Otherwise, not the end of the world.

vbrown608:
Sorry I've been slow the get to this. The subscribe effect for tweets is looks awesome! I really like that workflow. I also like how you created a less EFF-y translation string.

A couple general comments

- For logged in users, the "Call now" button is missing some styles - see screenshot. Not sure if that's part of this PR or a separate issue.

screenshot from 2016-10-12 12-20-09

- Tiny nit, but the bottom margin of the twitter e-mail signup button doesn't match the other submit buttons. Looks like that's defined for the e-mail tool on line 651 of action_page.css.scss. Maybe there could be a single rule for all - otherwise the twitter submit selector could be added there.

The other (inline) comments are just css style recommendations.

wioux:
Alrighty, fixed up all the margin issues and nixed the !important rule. It feels like the styles for action pages are a little messy. Perhaps I'll try to clean up a little in a follow up. I couldn't reproduce issue with the "Call now" button. :confused:

I also pushed a change to make sure `opt_in=true` is used for the CiviCRM subscriptions created by these actions.

mfb:
opt_in is false when we have a simple subscribe form so, by my interpretation of that logic, it needs to be false for these?

wioux:
Re-enabled the opt in for simple forms. I also found/fixed that call button styling issue you mentioned @vbrown608.

vbrown608:
:confetti_ball: :tada: :balloon: :champagne:

PR #218, opened 2016-10-04
Resize the window when the sidebar grows.

### LINKED ISSUE #125: action-content needs a minimum height
Currently, a short enough action description makes it impossible to complete the action for MOCs that have several form fields.

vbrown608:
Fixes #125.

Previously, when an action had a very short body and description, and users submitted their address in order to contact congress, the window would not grow large enough to show all the content inserted into the sidebar.

This pull request fixes that issue by triggering a resize when the content is inserted.


wioux:
Great. Fixes the same issue on email actions as well.

PR #221, opened 2016-10-07
Serialize updated_at timestamp more precisely

### LINKED ISSUE #220: Sometimes on save the app incorrectly warns me that I'm editing a stale version of an action

wioux:
This branch fixes #220. When a timestamp is used for a field value, `#to_s` is called which truncates the fractional seconds part. If the value is later parsed and compared against the original time it will appear to be in the past. This branch ensures the field value includes the fractional seconds part.


vbrown608:
Looks good!

PR #224, opened 2016-10-11
Filter action pages by issue, return json

vbrown608:


vbrown608:
Is this too much to put in the controller? 

Would it be better to add something like http://stackoverflow.com/a/21257685 on the model and chain it?

wioux:
The action isn't too complex IMO. If there were any other filter parameters I'd want a scope but with just one it seems ok.

We should customize the json response a bit, to include the feature photo url (right now it's just the file name), and also possibly render the markdown fields? I'm not certain that's required, I could see the client perhaps wanting to render with different options (like filtering html), but it's something to consider.

vbrown608:
Good point. I think rendering the markdown in rails is best in case there are things that only the action center will know about, like inline image paths.

wioux:
Sorry I didn't get to this yesterday... Looks good. I like custom json templates. :smile_cat: We also need the url for the action page and then this looks done.

PR #226, opened 2016-10-13
Action page issues

wioux:


vbrown608:
Cool - I had never seen Struct.new before, and loading the yaml on calls to .all seems like a good system!

The field title "Issue" reads a little weird (and EFF specific) to me. Maybe we should change it to topic? And if so, would it be worth it to change the name in code as well? 

I was also going to suggest making the select match the select element on the archive tab, but it looks like there are other non-matching selects as well. I'm going attempt a little refactor for a separate PR so we can hopefully get the right styles without adding class=form-control to everything all the time.

wioux:
There's a thing called "topics" used to pre-fill values in congress forms which I think is already easy to confuse with the thing we're adding here. I do think we could do better than Issue though... Category? Group?

mfb:
another generic term would be tags

vbrown608:
Oh yeah, I see the other "topic"...that's unfortunate. In that case I like Category.

PR #228, opened 2016-10-18
Replace yaml list of categories with database records

wioux:
For convenient creation of categories from the yaml file:

```ruby
YAML.load_file("config/categories.yml").each{ |category| Category.create(title: category) }
```


vbrown608:
Thanks for updating this! I think it would be nice to still conditionally display the categories select element based on whether there are any categories.

Issue #71, opened 2016-01-16
Show page for partners broken


TheNotary:
https://act.eff.org/admin/partners/5

There links to the pages on https://act.eff.org/admin/partners btw.

TheNotary:
The acts_as_paranoid gem is used to allow soft delete for Partner records.  It's mentioned to be buggy on rails 4.  It was always listed in the Gemfile to point to a github link rather than a gem w/ a version.  In rails 4, we can use concerns to handle this better and likely fix the bug I posted above.

PR #231, opened 2016-10-18
Fix partner link

vbrown608:
Just a lil bugfix


wioux:
The views from the old /admin/ namespaced partner show/edit pages are [still around](https://github.com/EFForg/action-center-platform/tree/master/app/views/admin/partners). Those should be safe to delete. Looks like this fixes #71.

app/views/admin/partners/index.html.erb
```diff
@@ -32,7 +32,7 @@
                   <%= link_to partner.name, partner -%>
                 </td>
                 <td> <%= partner.code -%> </td>
-                <td> <%= link_to 'Link', partner.privacy_url -%> </td>
+                <td> <%= link_to 'Link', partner.privacy_url if partner.privacy_url-%> </td>
```
wioux: Can we check for partner.privacy_url.present? here? Or, otherwise, normalize the field to be non-empty. Partner#privacy_url can end up as "" as a result of editing.

PR #234, opened 2016-10-19
No sunlight

wioux:
Roll our own Congress member store/lookup and nix sunlight-congress. Replaces the half baked `Congress` module with a `CongressMember` model. Also support call-congress's new LOCATION_DISTRICT campaign type (smarty streets returns congressional district so we can use that to look up members).


vbrown608:
This is really nice!! Do you know what the code in 77a33d2 was supposed to do? 

:sun_behind_small_cloud: :sun_behind_large_cloud: :cloud:

wioux:
Ah that's one thing I was a little uncertain about.. There is a model `CongressScorecard` which counts the number of constituents that have taken a particular action for each congress member, and we had an action at `/admin/petitions/:id/reports` which would show you a congress member's count for a petition as well as a list of signatures from their state.

I'm assuming it's a dead feature since there wasn't any similar report for email campaigns (despite their also keeping a score card), and there is no way to access the report action by following links on the site.

vbrown608:
Okay, let's go ahead and we can add it back in if activism asks for it.

PR #235, opened 2016-10-20
Better privacy for user registration

vbrown608:
Resolves redmine #11111

### DIFF
--- a/app/views/user_mailer/signup_attempt_with_existing_email.text.erb
+++ b/app/views/user_mailer/signup_attempt_with_existing_email.text.erb
@@ -2,7 +2,7 @@ Hello <%= @user.email %>!
 
 Someone tried to sign up for a new Action Center account with this e-mail address, but an account already exists. If you forgot your password, you can change it with the link below.
 
-<%= link_to 'Change my password', edit_password_url(@user, :reset_password_token => @token) %>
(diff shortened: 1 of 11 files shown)


app/views/user_mailer/signup_attempt_with_existing_email.text.erb
```diff
@@ -0,0 +1,8 @@
+Hello <%= @user.email %>!
+
+Someone tried to sign up for a new Action Center account with this e-mail address, but an account already exists. If you forgot your password, you can change it with the link below.
+
+<%= link_to 'Change my password', edit_password_url(@user, :reset_password_token => @token) %>
```
wioux: The `link_to` here creates an HTML tag in our plain text format.

vbrown608:
Read about `yield resource if block_given?` and realized there's a much better way. Wish I'd seen that sooner!

Thanks for catching those differences between the real success state and the fake success state. I kept the change to `after_update_path_for` because I think the pending confirmation message was getting a bit lost before.

wioux:
Yeah that yield is sweet. It looks like we still need to do something if the email is taken and there are separate validation errors as well?

wioux:
:+1: :shipit:

PR #239, opened 2016-10-25
Remove some comments/docs/privacy notes about sunlight api

wioux:
This branch removes some references to the app's use of the Congress API.


vbrown608:
:100:

Issue #27, opened 2015-10-16
Implement Roles


TheNotary:
As it stands, there are two types of users, admins and non-admins.  Below is ideal:

```
Roles:
-activism (admin as it is now, can edit ActionPages)
-administration (can grant other users the role of activism or partner)
-partner (can download CSVs and define news letter privacy policies, etc)
```

TheNotary:
Roles goes hand in hand with authorization.  I'm building a branch in which I propose using the gems `cancancan` and `rolify` to suffice our role authorization.  The branch will include: 

* Gem inclusions
* Definitions of roles in-code
* Specifications of roles onto users
* minor re-writes of existing authorization code

Here's an example of some existing authorization code:

(app/controllers/action_page_controller.rb)
```
#set_action_display_variables
.
.
.
# Redirect visitors to archived actions unless they have taken that action.
if @actionPage.archived? and @actionPage.archived_redirect_action_page_id and !@actionPage.victory?
  taken_action = false
  unless current_user.nil?
    taken_action = true if current_user.events.actions.where(action_page_id: @actionPage).first
  end
  return redirect_to(action_page_path(@actionPage.archived_redirect_action_page_id)) unless taken_action || current_user.try(:admin?)
end
```

Before I do roles, I'd like to write a cucumber test so it's easy to check off that my changes are valid non-breaking changes.

elliotharmon:
Is the partner role site-wide, or specific to a campaign? If it's site-wide, I worry that there might be a privacy policy issue in giving them ability to download CSVs.

mfb:
@elliotharmon I assume this is referring to the existing concept of partners in the app: they can download CSVs for their organization.

elliotharmon:
Thanks @mfb. I didn't realize that existed already. Ignore my comment.

PR #242, opened 2016-10-26
Collaborator role

wioux:
This PR adds the collaborator role (see #27). Collaborators get limited access to the admin area:
- They can browse the action pages list
- They can view draft actions
- They can view site and action page analytics
- They can view and delete petition signatures

I use a simple helper to assist with whether to show content to a user or not based on role permissions, like so:

``` html
  <% if can?(:view_analytics) %>
    <%= render "panel_analytics" %>
  <% end %>
```

Backed by this method on `User`:

``` ruby
  def can?(ability)
    case ability
    when :browse_actions
      admin? || activist? || collaborator?
    when :administer_actions
      admin? || activist?
    when :administer_homepage
      admin? || activist?
    when :view_analytics
      admin? || activist? || collaborator?
    when :administer_partners?
      admin? || activist?
    when :administer_topics?
      admin? || activist?
    when :administer_users?
      admin?
    else
      admin?
    end
  end
```

I think this should cover our needs and that we can avoid bringing in cancancan for now.

A new tab is added to the admin area that allows listing/finding users and granting them collaborator access.


wioux:
Fixed a merge conflict. Also, I got confirmation today that it was meant to be a site wide role.

vbrown608:
Thanks for following up with activism. 

It looks like the "add as collaborator" button might not be updating user records.

Issue #175, opened 2016-09-12
Javascript cleanup (more)


wioux:
Separate issues from #65 
- [x] Clean out inline `<script>` tags
- [x] Clean out global variables like [date_start and date_end](https://github.com/EFForg/action-center-platform/blob/master/app/views/admin/action_pages/_date_vars.html.erb), [petition_id](https://github.com/EFForg/action-center-platform/blob/master/app/views/tools/_petition.html.erb#L2), [welcome.js stuff](https://github.com/EFForg/action-center-platform/blob/master/app/assets/javascripts/application/welcome.js), etc.
- [ ] Standardize on templating framework (ejs and underscore are both used). Remove unnecessary templates, preferring rendered html.
- [x] Move third party js under vendor/assets

k-stewart:
@wioux Should this be closed?

PR #244, opened 2016-10-28
Javascript cleanup

wioux:
This branch is mostly moving things out of global scope / into the document.ready callback. I also moved congress-forms.js under vendor/. It addresses a few points from #175.


app/assets/javascripts/application/sticky_tools.js
```diff
@@ -55,6 +57,3 @@ $(function() {
     scroller();
   }
 });
-
```
vbrown608: This is being used to make the save/view/preview/csv buttons sticky on edit action pages (using `data-spy="affix"`). We should probably include it somewhere better, or I'd be open to getting rid of that functionality - just wanted to flag that it would be going away. 

I tried to use it for the actual sticky tools so they could share one solution, but I think the sticky tools need something fancier.

vbrown608:
Thanks for cleaning this up! Aside for that one inline comment, I think this ready for merge.

app/assets/javascripts/application/sticky_tools.js
wioux: I replaced it with `//= require bootstrap-sass/bootstrap/affix`. This seems to be just the minified version of that file inlined for some reason.

PR #249, opened 2016-11-03
Fix infinite loop on password reset

vbrown608:
When user.password_expired was set to false (rather than true or nil), resetting the password would result in an infinite loop. 

This commit fixes that bug and preserves the behavior where resetting a password sets password_expired to false.

Resolves 11301 in Redmine (both Starchy and the submitter have password_expired=false, while most users have password_expired=nil)


wioux:
:+1: this was annoying

PR #250, opened 2016-11-03
Fix cron/delayed_job worker containers

wioux:
* We shouldn't precompile assets in cron/delayed_job containers
* cron runs jobs in a blank environment, so we need to somehow re-source all our variables before calling `bundle exec rake`. I did this by exporting the environment to /root/.profile in entrypoint, but I'm open to feedback on better approaches...


docker/entrypoint.sh
```diff
@@ -22,4 +22,6 @@ fi
 # host ip needed by application for better_errors whitelisting to work
 export HOST_IP=`/sbin/ip route|awk '/default/ { print $3 }'`
 
+printenv | sed 's/^\([[:alnum:]_]*\)=\(.*\)$/export \1="\2"/' >/root/.profile
```
vbrown608: :dizzy_face:

PR #252, opened 2016-11-04
Take docker environment from .env

### LINKED ISSUE #248: docker-compose.yml and .env
docker-compose.yml should load the environment from .env. We can make .env.example to hold the expected keys. This will also be slightly more useful for those developing/deploying without docker.

wioux:
Closes  #248. Here's a command to generate .env from docker-compose.yml:

```bash
$ ruby -ryaml -e 'YAML.load_file("docker-compose.yml").dig("services", "app", "environment").each{ |k,v| puts %(#{k}="#{v}") }' >.env
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,83 @@
+#
+# Start of application environment variables
+#
+
+# Standard Rails stuff (required)
+secret_key_base=
+devise_secret_key=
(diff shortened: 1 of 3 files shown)


vbrown608:
yay best practices! sweet yaml magic

PR #256, opened 2016-11-07
Set Docker user to www-data

vbrown608:
- Resolves #254 by setting the user to www-data in the Dockerfile.
- Logs to stdout in production, which is a best practice for Docker.
- Changes are still needed to make this Dockerfile work with the cron service.


wioux:
Looks good. I will look at the cron environment issue in a follow up branch.

Issue #255, opened 2016-11-07
Action page admin improvements


wioux:
- Admin area
  - [ ] If it's only possible to have 1 action type per page, the form should enforce that (e.g. radios instead of check boxes)
  - Calls
    - [ ] Integrate with Call Power to allow selecting campaign by title
  - Petitions
    - [ ] Clarify that "Goal" is a number of signatures and is required
  - Tweet
    - [ ] UX for adding targets by handle should be better. Instead of a textarea with one target per line (without the @) there should be a regular input which adds targets to the handle table immediately.
  - Email
    - [ ] Option to target specific congress members instead of specific bioguide IDs.
    - [ ] The UI should enforce that you cannot target both congress and specific email addresses at the same time.
    - [ ] Possibly the "send message to congress" actions should not even be described as "email" but be a separate action type altogether. In any case, the user shouldn't see form elements for "Topic Category" / "Campaign tag" when setting up a regular email campaign.

vbrown608:
Admin area: It's currently possible to have multiple actions per page, and some action pages do.

Calls: That would be awesome.

Petitions: I don't think goal is required - without it, users will see a signature count rather than a progress bar. This is particularly helpful in campaigns like reclaim invention where a single goal doesn't necessarily make sense.

Email: I agree that this is a confusing name for submitting comments to congress.

lenazun:
Feels like some of these will be tackled during admin redesing

Issue #193, opened 2016-09-20
Help text + documentation for call tool integration


vbrown608:
Right now, there aren't a lot of cues to help a non-EFFer understand how to integrate with the call tool. I suggest:

- [ ] A page in the repo wiki about adding the call tool API key to your app, finding a call campaign id, and additing it to an action

- [ ] Minimal help text on the admin/action settings panel - perhaps when hovering over "Campaign ID", similar to the hover text about photo specs for tweet actions

PR #258, opened 2016-11-08
Call power integrate campaign select

wioux:
Part of ideas from #255.


vbrown608:
Awesome, I'll look at this soon. I think it will also address item 2 of #193 really nicely.

PR #260, opened 2016-11-08
Cron worker fix

wioux:
Follow up on #256.


vbrown608:
cool, user setting in docker-compose is handy! Didn't know you could do that.

PR #261, opened 2016-11-08
Assets precompile

### LINKED ISSUE #185: Find a happy home for rake assets:precompile

vbrown608:
- Remove remaining environment-based config from javascript assets
- Move assets precompile and webshims back to the dockerfile (thanks to changes to recommended mounted volumes in docker-compose.example.yml)

Resolves #185


wioux:
Good riddance #185.

PR #265, opened 2016-11-16
Topics tab changes

wioux:
This branch introduces UI changes to the "Topics" admin area. It's a follow up to #262 which introduced tests for that tab.

The UI is largely the same except that instead of modal dialogs for adding/deleting topics/categories there are instead inline forms. It uses Reactjs and removes backbone from the app.

I've also renamed the tab to "Congress Topics".


vbrown608:
This looks great! Yay for eliminating unnecessary modals. Nice improvement to the path for this page too :)

A few comments:
- Any luck tracking down the source of the weird indent on some categories? (ex TPP in the staging database)?
- Something weird happens when I resize to mobile-width and then expend back to normal width.
- imo a visible label on the "Add topic" element would be helpful (ie removing the sr-only class). I would change it to "Add term" to match the column header.
- Should the "Create new topic category" button still be there? If it's replaced by the "Category name" + Create widget at the bottom:
  - I think the "Category name" field should match the text size and padding of the topic fields
  - I suggest moving that widget to the top of the category list, and inserting new categories at the top. That way users can still find it easily if there are many categories.

vbrown608:
I guess "Category name" does match the actual category names in terms of style. I'd be fine with matching it to the other (smaller fields) or just cleaning up the padding a bit so it aligns with Create button and has some left padding within the field.

wioux:
Thanks for the feedback. This branch was a little unpolished... All this stuff is fixed now. There was a bug in bootstrap-responsive-tabs causing that panel collapse issue so I updated that library.

vbrown608:
So many little pieces all lined up (':

I think it might be helpful to insert new topic categories at the top where users can see them. I went ahead and made that change to understand the code a little better. Feel free to merge if that looks good to you, or revert/make changes if a different approach is better

PR #266, opened 2016-11-29
Do ahoy visit tracking via image tag embedded in page

wioux:
Our action view tracking is currently problematic since those pages are cached behind a proxy. In this branch I instead create the ahoy event out of an embedded image on the action page. We can cache the action page and configure the proxy to always pass /ahoy/visit.gif through to the app.

### DIFF
--- a/app/controllers/action_page_controller.rb
+++ b/app/controllers/action_page_controller.rb
@@ -139,17 +139,6 @@ private
                               country_code: current_country_code,
                               email: current_email }
 
-    # Tracking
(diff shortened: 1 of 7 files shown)


mfb:
FYI we already have anon-stats.eff.org (piwik) tracking page views (and other user actions) via image or ajax request.  Probably easier to store redundant data in the rails db, but we can also pull stats via piwik API

vbrown608:
Nice solution.

PR #268, opened 2016-12-05
Congress messages

wioux:
This branch creates a new model and admin UI for the the "send a message to congress" actions which is separate from those of the regular email actions, fixing some issues from #255. The name for the new model is CongressMessage, though that may be subject to change...

It includes [a migration](https://github.com/EFForg/action-center-platform/blob/congress_messages/db/migrate/20161109191337_create_congress_message_campaigns.rb) for replacing `email_campaigns` that target either House or Senate with new `congress_message_campaigns` that do the same, while leaving the campaigns that target specific email addresses as they are. It also alters all the `ahoy_events` related to sent messages, which takes a few minutes. The migration is reversible and must be rolled back if the branch is reverted.


wioux:
Found an annoying merge blocker -- the /account page lists all actions with an icon for each type. These icons are taken from a different set than the ones used in the admin area. It's a much more restricted [choice](https://github.com/EFForg/action-center-platform/blob/master/vendor/assets/stylesheets/eff-icon-font.css.scss#L90-L121). The only one that looks workable is icon-pencil.

The underlying font is aggregated from [various other icon sets](https://github.com/EFForg/action-center-platform/blob/master/vendor/assets/fonts/config.json). I'm going to see what's involved in extending it.

wioux:
Solution for now is to bring ion-icons out of the admin area and use them on the user profile as well.

vbrown608:
Nice migration! Few comments on frontendy things:

- The settings pane for congress messages is missing the little up/down arrow that the other have
- The labels for special text for congress messages ("Replace "Email Your Representatives" text with:" etc) are closer to the preceding field than the field they label, which is a bit confusing. 
- The autocomplete for bioguide ids is a great idea
  - I think we could update the placeholder text to something like "Start typing a legislator's name or bioguide id". That way users won't go look up the bioguide id only to realize it wasn't necessary.
  - I'm using select2 for similar functionality on act.eff.org/action/reclaim-invention. It might be good to go with one or the other. I wasn't totally happy with select2 but I think there are pros and cons to each...maybe we can talk about it.

wioux:
Thanks. I've switched it to select2 and in addition to eliminating a redundant library I think the UI is much improved as a result.

PR #269, opened 2016-12-06
Only allow call actions if call tool credentials are provided     

### LINKED ISSUE #259: Only allow call actions if call tool credentials are provided
Right now not having the call tool set up causes parts of the action center to 500. I think it would be nice to disable those parts of the site if call tool credentials aren't provided. Here's one idea:

- In the "edit action page" area, the "edit call action" pane could show some info about setting up the call tool. 
  - The checkbox would be shown as disabled.
  - The help text should also look obviously disabled - maybe black on a light grey background.
  - The config for that call action should be saved so that it will be automatically reenabled if credentials are added.
- If an action page already exists with a call action, the tool container would be hidden on that page

wioux:
Branch to fix #259

### DIFF
--- a/app/assets/javascripts/admin.js
+++ b/app/assets/javascripts/admin.js
@@ -17,7 +17,8 @@
 $(document).on('ready', function() {
 
   var editor = initEpicEditor('action-page-description', 'description');
-  var editor2 = initEpicEditor('action-page-what-to-say', 'what-to-say');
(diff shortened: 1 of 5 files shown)


vbrown608:
I like this! Everything reads really clearly as disabled to me and the warning for admin on action pages is great.

PR #271, opened 2016-12-09
Add screen-reader accessible labels to form fields

vbrown608:
Many of the form fields in the action center rely on placeholder text, which isn't typically read out by screen readers. I think the best UX would be to have visible label elements for all fields. In the meantime I added aria-labels so that users with screen readers can use the action center.

Fixes #271


wioux:
Looks good. :+1:

PR #272, opened 2017-01-07
Allow embedding form authenticity tokens in <esi:include> tags

wioux:
Related: https://www.fastly.com/blog/caching-uncacheable-csrf-security


vbrown608:
Cool!

PR #276, opened 2017-01-12
Don't poll updated at

wioux:
This branch changes the approach to concurrent modification avoidance from polling the record's updated_at timestamp to checking it once on save. It's slightly more accurate, since the polling happened once per 10 seconds, but moreover keeps the logs cleaner and reduces load on the server.


vbrown608:
This seems like a much better way to handle it. 

It looks like you're removing the updated_at route but not the controller action.

wioux:
Good catch, thanks.

PR #277, opened 2017-01-19
Remove dead/unnecessary cookies and flash notices

wioux:
This branch removes some references to `request.session` which should be dead anyway. This is true except for one `flash.notice` that was removed -- but it is only displayed to no-js users, and it's clear from the rest of the page that the petition was signed.

As of this branch, I believe the only parts of the app that depend on cookies are either only accessible when logged in, or in response to a POST request, or in certain restricted controllers (`AhoyController`, `CsrfProtectionController`) . After it's merged we can enable some Fastly features that we couldn't use before.


wioux:
Fixed the failing test by rebuilding. It seems to fail intermittently, about half the time, and is unrelated to this branch... I've opened an issue about it #278.

PR #283, opened 2017-03-06
Fix rep lookup error message

### LINKED ISSUE #282: Display error if representative lookup fails
It seems currently the form just stalls out (even though the backend reports an error).

wioux:
Closes #282


wioux:
This is triggered in the case where `/tools/reps_raw` returns an error.

vbrown608:
Looks good!

PR #285, opened 2017-03-14
Verify authenticity token for petition actions

vbrown608:
I'm not sure why this was originally added (git history doesn't go that far back) but I think we can remove it now.


vbrown608:
Closing in favor of #396

PR #287, opened 2017-03-27
Don't post to call power if district lookup failed

wioux:


mfb:
Are there valid district zero that evaluate to false? for at-large districts?

wioux:
At-large is given as "AL" (https://smartystreets.com/docs/cloud/us-street-api#congressionaldistrict).

I'm going to revisit this area soon to add UI feedback for a failed lookup.

PR #291, opened 2017-04-11
Don't message defunct phantom-dc legislators

wioux:


app/assets/javascripts/application/tools/congress_message.js
```diff
@@ -247,6 +247,18 @@ $(document).on("ready", function() {
             },
             error: function() {}
           });
+        },
+        onDefunctLegislator: function(bioguide, contact_url) {
+          var notice = $("<p>").addClass("defunct-notice")
```
otherjon: Do we support blank contact_urls?  I thought that was allowed, since they're optional in the rake tasks.

wioux: I've altered the message to take into account a possible null contact_url.

Issue #293, opened 2017-05-01
Resend unlock instructions page is unstyled


williamtheaker:
https://act.eff.org/unlock/new

PR #294, opened 2017-05-01
Style Resend unlock instructions page

williamtheaker:
Untested fix for #293

### DIFF
--- a/app/views/devise/unlocks/new.html.erb
+++ b/app/views/devise/unlocks/new.html.erb
@@ -1,12 +1,20 @@
-<h2>Resend unlock instructions</h2>
+<div class="redtop">
+  <div class="container">
+    <div class="row">
+      <div class="resend-unlock col-lg-4 col-lg-offset-4 col-sm-6 col-sm-offset-3">
+        <h2>Resend unlock instructions</h2>
 
-<%= form_for(resource, :as => resource_name, :url => unlock_path(resource_name), :html => { :method => :post }) do |f| %>
-  <%= devise_error_messages! %>
+        <%= form_for(resource, :as => resource_name, :url => unlock_path(resource_name), :html => { :method => :post }) do |f| %>
+          <%= devise_error_messages! %>
 
-  <div><%= f.label :email %><br />
-  <%= f.email_field :email, :autofocus => true %></div>
+          <div><%= f.label :email %><br />
+          <%= f.email_field :email, :autofocus => true %></div>
 
-  <div><%= f.submit "Resend unlock instructions" %></div>
-<% end %>
+          <div><%= f.submit "Resend unlock instructions" %></div>
+        <% end %>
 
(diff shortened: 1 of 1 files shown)


app/views/devise/unlocks/new.html.erb
```diff
@@ -1,12 +1,20 @@
-<h2>Resend unlock instructions</h2>
+<div class="redtop">
+  <div class="container">
+    <div class="row">
+      <div class="resend-unlock col-lg-4 col-lg-offset-4 col-sm-6 col-sm-offset-3">
```
wioux: Can we change the columns to `col-lg-6 col-lg-offset-3 col-sm-8 col-sm-offset-2`?

app/views/devise/unlocks/new.html.erb
```diff
@@ -1,12 +1,20 @@
...
+        <h2>Resend unlock instructions</h2>
 
-<%= form_for(resource, :as => resource_name, :url => unlock_path(resource_name), :html => { :method => :post }) do |f| %>
-  <%= devise_error_messages! %>
+        <%= form_for(resource, :as => resource_name, :url => unlock_path(resource_name), :html => { :method => :post }) do |f| %>
+          <%= devise_error_messages! %>
 
-  <div><%= f.label :email %><br />
-  <%= f.email_field :email, :autofocus => true %></div>
+          <div><%= f.label :email %><br />
```
wioux: Add `class="form-group"` to the `<div>` and `class: "form-control"` as an option to the email field.

app/views/devise/unlocks/new.html.erb
```diff
@@ -1,12 +1,20 @@
...
+          <%= f.email_field :email, :autofocus => true %></div>
 
-  <div><%= f.submit "Resend unlock instructions" %></div>
-<% end %>
+          <div><%= f.submit "Resend unlock instructions" %></div>
```
wioux: Adding `class: "btn btn-default"` as an option to the submit improves its styling.

app/views/devise/unlocks/new.html.erb
```diff
@@ -1,12 +1,20 @@
...
+        <% end %>
 
-<%= render "devise/shared/links" %>
+        <%= render "devise/shared/links" %>
+      </div>
+    </div>
+  </div>
```
wioux: This is kind of silly but can you drop this in under the close of the `<div class="container">` element?

```svg
  <svg class="triangle" viewBox="0 0 24 23.8" preserveAspectRatio="none">
    <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#triangle"></use>
  </svg>
```

We should really have a partial or helper for the triangle, but it's hardcoded everywhere else..

wioux:
Thanks @williamtheaker! I made some notes in places that could benefit from bootstrap classes.

PR #298, opened 2017-06-13
Validate atom feed.

### LINKED ISSUE #297: rss feed doesn't validate
One of our user writes

> https://act.eff.org/action.atom
> I use an RSS aggregator on my site and this particular feed does not work.
> If you go to Feed Validator you can see why it doesn't work.

wioux:
Fixes #297


mfb:
looks good to me. Ideally there would be a test to confirm feed is valid, but in practice that might not be easy to do aside from submitting it to the feedvalidator website.

PR #303, opened 2017-07-21
Save SES complaint notifications

vbrown608:
Creates an endpoint at /complaint to log notifications of SES complaints.

This seems to work correctly when I sign a petition on act.staging.eff.org as  complaint@simulator.amazonses.com

When we're ready to use this in production, we can add another subscription to the SNS topic associated with our ses compaints ("ses-complaint"). We confirm that subscription by visiting a url sent to the subscribed endpoint. For that reason, I'm logging requests to the endpoint even if they don't conform that to the complaint format.

I pulled out a few fields from the complaint that seemed useful, and also stored the rest of the complaint as a weird postgres json field thing in case we want to get more info out of it.


mfb:
I think it would it be better to store as jsonb? That way we can index the json columns, like MessageID, and do fast lookups.   (Or if we have an ancient postgres that doesn't do jsonb we could make message ID be a column.)

Perhaps we could also create an info-level log entry after a notification is saved? That way someone looking at logs has a chance to notice a high frequency of complaint notifications coming in.

vbrown608:
Sounds good!

app/controllers/sns_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def bounce
+    begin
+      success = true
+      message = JSON.parse(request.body.read)['Message']
+      bounce_arr = JSON.parse(message)['bounce']['bouncedRecipients']
+      bounce_arr.each do |recipient|
+        Bounce.create(email: recipient['emailAddress'].downcase)
+      end
+    rescue
```
wioux: What's this meant to rescue?

wioux: I think if we check for `message["notificationType"] == "Bounce"` then we shouldn't need begin/rescue/ensure?

app/controllers/sns_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def complaint
+    logger.info 'Received Amazon SES complaint notification'
+    begin
+      success = true
+      message = JSON.parse(request.body.read)['Message']
+      complaint = JSON.parse(message)['complaint']
+      if complaint.nil?
+        # Save non-standard requests, ex subscription confirmation
```
wioux: Are subscription confirmations expected at the /complaint endpoint? Why do we save them?

app/controllers/sns_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+    rescue
+      success = false
+    ensure
+      render json: {success: success}
+    end
+  end
+
+  def complaint
+    logger.info 'Received Amazon SES complaint notification'
```
wioux: Can we send a message to Sentry as well?

```ruby
Raven.capture_message("Received Amazon SES complaint notification", level: "info")
```

vbrown608: Good idea

app/controllers/sns_controller.rb
vbrown608: Yeah, they go to the same endpoint that the complaint notifications will eventually go to. 

I was thinking if the format changes for some reason it might be nice to still capture what amazon sends us - posts from anyone else won't make it to this controller because they won't have the aws key. I suppose the alternative would be to just log the confirmation url - there should only be one.

app/controllers/sns_controller.rb
vbrown608: Yeah, I based this on the old code but that seems better. It was hard to find errors when they kept getting rescued.

vbrown608: Do you think it's reasonable to expect that the message from Amazon will conform to the formats they've described and let the code error if it doesn't? Or would it be better to check any possible errors based on unexpected format, ex. check that `message` isn't empty before the second call to `JSON.parse`?

wioux: I think an exception is appropriate if the input is unexpected. Lately I've been using Sentry event contexts and finding them really useful. We could do something like this

```ruby
  before :set_context

  def set_context
    Raven.extra_context(message: request.body.read)
    request.body.rewind
  end
```

Then if there's a parse error we'll get notified and also have the post body available to inspect.

app/controllers/sns_controller.rb
wioux: Is it possible to get them sent to a different route? It seems random that subscription confirmations show up at POST /complaint.

vbrown608: No, I didn't see a way. Unfortunately there's not much to configure. It sounds like maybe if we set up contexts as you're describing above, though, we can just pull the URL out of sentry.

vbrown608: (I think they basically want to see that you control the endpoint you're asking them to send stuff to)

wioux: Ok. This makes sense now. Sorry, should have read the SNS docs more carefully.

vbrown608:
Thanks for the feedback Peter! I think this looks nicer. 

I pushed f73e66a after reading this: http://craftingruby.com/posts/2015/05/31/dont-use-before-action-to-load-data.html. Curious if you agree.

wioux:
+1 looks good to me. I agree, I like controller actions that are "complete" in that sense.

PR #319, opened 2017-09-06
Add partners to congress forms

### LINKED ISSUE #311: Allow multiple partners on an action

### LINKED ISSUE #313: Add to partners to action pages, rather than specific tools w/in the action

vbrown608:
This PR does a couple things:
- Allows multiple partners email newsletter signups on a single action
  - Partners are now joined to action pages using the "partnership" model
  - Users completing an action see EFF's signup alongside a randomly chosen partner
- Gives admin the ability to put a partner signup form on petition and congress message actions, and begins adding it to the other action types

A few things still todo:
- I'd like generalize processing the partner email signup for all actions and add tests for that. In this PR the congress message action just duplicates a version of the logic petition action.
- There's some inconsistency in the app around whether logged in users should be prompted to subscribe to EFF's mailing list. They probably always shouldn't, but I need to follow up, and then I can centralize that logic somewhere.
- I'm not sure about the best way to add templated content (such as the email signup radios) to the second phase of the congress forms action. In this PR I'm rendering it into a hidden div and then injecting it into the form later.

This resolves #311 and partially [not using the R word so it doesn't get closed] #313.


app/controllers/action_page_controller.rb
```diff
@@ -109,6 +109,7 @@ def set_action_display_variables
     @tweet = @actionPage.tweet
     @email_campaign = @actionPage.email_campaign
     @congress_message_campaign = @actionPage.congress_message_campaign
+    @partner = @actionPage.random_partner
```
wioux: This is choice is going to cached for some time behind Varnish. We should follow up and move the partner selection to javascript.

app/views/tools/_petition.html.erb
```diff
@@ -124,10 +124,7 @@
               <% unless @require_location %>
                 <%= render 'tools/petition_location', f: f %>
               <% end %>
-
-              <%= render "tools/email_signup_radios" unless user_signed_in? %>
```
wioux: Non-blocking: Is it intentional that the `unless user_signed_in?` condition is lost?

wioux:
I see no blockers. :+1:

vbrown608:
Thanks for the rush review!! I've you've got any other feedback or thoughts, I'll definitely make time for a follow-up.

app/models/action_page.rb
```diff
@@ -80,4 +81,8 @@ def image
   def no_drafts_on_homepage
     FeaturedActionPage.where(action_page_id: id).destroy_all unless published?
   end
+
+  def random_partner
+    partners.where(id: partners.pluck(:id).sample(1)).first
```
wioux: `partners.sample` would have worked here, although style wise in that case `#random_partner` might be too simple to warrant its own method.

app/controllers/admin/action_pages_controller.rb
```diff
@@ -51,7 +51,7 @@ def create
   end
 
   def edit
-    @actionPage.partner ||= Partner.new
+    @actionPage.partners ||= Partners.all
```
wioux: I'm not sure this does anything since an uninitialized `has_many` relation is still non-nil (it's an empty ActiveRecord::Associations::CollectionProxy).

app/controllers/admin/action_pages_controller.rb
```diff
@@ -28,7 +28,7 @@ def update_featured_pages
 
   def new
     @actionPage = ActionPage.new
-    @actionPage.partner = Partner.new
+    @actionPage.partners = Partner.all
```
wioux: This should be fixed relatively soon -- new action pages are initialized with partnerships to everyone.

Issue #314, opened 2017-08-31
Provide partners with a link to their action that always includes their signup form


vbrown608:
When actions have multiple partners and display a random partner signup form, partners should have access to a special link, ex act.eff.org/action/my-action?partner=ff, that always shows their signup form.

vbrown608:
(This is lower priority than the other partners-related changes)

vbrown608:
Partners can now use act.eff.org/action/my-action?partner=[CODE]

PR #321, opened 2017-09-11
Show all partners on tools, process partners on call tool backend

vbrown608:
* Show all partners on tools and process all tools on the backend. This is a step towards randomly selecting a partner on the frontend - I wanted to render and process them all correctly so this will work for no-js users.

* Process partner submissions for the call tool.

* Stop using a special param to do determine if we should show the EFF email signup. We can handle this using the partner param in #314.

* Add an end-to-end integration test for congress forms, using the puffing billy gem to mock ajax calls to the contact congress service.


wioux:
The styling on the call tool radios is off. There is the an issue with that on production as well (#322), but perhaps we could fix it in this PR?

app/views/tools/_call.html.erb
```diff
@@ -119,10 +119,8 @@
           </div>
 
           <p style="text-align: center;">
-            <% if ask_for_newsletter_signup? %>
             <br />
             <strong><a href="/register">Sign up</a> for an EFF Action account and save your information for next time.</strong>
```
wioux: It still makes sense to hide this for logged-in users.

features/step_definitions/congress_steps.rb
```diff
@@ -0,0 +1,31 @@
+Given(/^a congress member exists$/) do
+  member = FactoryGirl.create(:congress_member)
+end
+
+Given(/^a stubbed phantomdc$/) do
+  Rails.application.config.congress_forms_url = "http://phantomdc.test"
```
wioux: Awesome! Though it further confuses me about how exactly cucumber/capybara works...

wioux:
This looks pretty good to me. I made a few notes but nothing is really blocking.  :+1:

app/views/tools/_call.html.erb
vbrown608: Good point

PR #325, opened 2017-09-12
Use rails-generated url for image in sass

vbrown608:
Original image path was causing 404 on production.


vbrown608:
This works on staging :+1:

Issue #323, opened 2017-09-12
Email button breaks container


vbrown608:
Perhaps we could remove the letter icon.

![mail-button-bug](https://user-images.githubusercontent.com/1887071/30303907-7a15d9a6-971f-11e7-9aaa-5611ad6ff763.png)

hughillustration:
Let’s! It’s ugly.

H.

PR #327, opened 2017-09-18
Add partners to email actions, post email action form directly to tools controller, show a random partner

vbrown608:
- Makes email actions work a bit more like other actions:
  - Email actions are now a form that posts to /tools/email (rather than using javascript to GET tools/email_target and trigger a form submission to the submissions controller)
  - The old email email controller was unused, I merged it with email_target
- Shows users a random partner signup on the client side so the partner choice isn't cached
- Overrides the CiviCRM::post method in tests so that submissions to civi are stored in memory rather than posted and tests those submissions
- Removes commented out and dead code for launching Yahoo mail :skull: 
- Removes the email icon from email tools, closing #323


wioux:
When you are logged in and there are no signup options (no partners), you still get asked for your email address:

email-action-logged-in-email-prompt

wioux:
There is an error when you leave `Email (optional)` blank but check yes for mailings. Email could dynamically become/unbecome required when you change the radio state?

vbrown608:
Good idea.

wioux:
:+1: good to go

PR #345, opened 2017-12-08
Bot prevention

vbrown608:
Attempt to reduce successful registrations by bots by doing two things:
- rejecting form submissions where a hidden field is filled out. this returns a 200 status and no content.
- rejecting form submissions that take fewer than four seconds. this displays a descriptive error and reloads the fields.

Banning IPs that fill the hidden field is also an option.


wioux:
Interesting approach. :+1: once tests are happy.

PR #346, opened 2017-12-18
Throttle password reset

vbrown608:
For https://redmine.eff.org/issues/14947

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,7 +1,7 @@
 class Rack::Attack
   throttle('registrations', limit: 10, period: 1.day) do |req|
     if req.path == '/' && req.post?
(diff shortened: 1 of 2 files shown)


config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,17 @@
+class Rack::Attack
+  throttle('registrations', limit: 10, period: 1.day) do |req|
+    if req.path == '/' && req.post?
+      if req.env['HTTP_X_FORWARDED_FOR']
+        req.env['HTTP_X_FORWARDED_FOR'].ip.split(/\s*,\s*/)[0]
```
wioux: The `.ip` in the middle there looks like a mistake.

vbrown608: oops yep thanks

PR #347, opened 2017-12-19
Quieter production logging

vbrown608:
This stops logging of IPs in production and also makes production logging much quieter. The biggest change is that it doesn't log rendering of view partials. The new format looks like
```
method=GET path=/ format=html controller=WelcomeController action=index status=200 duration=154.53 view=17.02 db=71.27
```
I think quieter production logs sound good, @wioux lmk if you are finding the other information useful.


vbrown608:
(I marked this WIP so I can test it with Puma)

PR #350, opened 2018-01-02
Use Puma instead of Unicorn to serve Rails

vbrown608:
This depends on #348 (Gemfile.lock updates some codeclimate dependences causing Travis to fail, but #348 removes codeclimate entirely)

We will need remove the cmd override from the production docker-compose file before we deploy this.


vbrown608:
Merged that other PR, this is ready for review now.

PR #354, opened 2018-01-26
Store sessions in the database

vbrown608:
This PR moves session storage to the backend in response to Redmine #14934. 

It creates a database record for ever unique visitor to the site. Currently we already create database records for each action page view using Ahoy, but we should be aware of this in case we run into performance issues. 

I added a cron job to trim sessions for the database that are over a year old. I'm not sure what the session duration is for the action center - it looks like we're using Devise's default of 2 weeks, but that seems surprisingly short, so I'm checking that I actually get logged out after that time frame. If that's the case we should make it longer, and either way it should match the amount of time we keep sessions around in the database.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -80,6 +80,7 @@ sentry_dsn=
 
 # Miscellaneous (not required)
 # set to true to use <esi:include> tags in place of form authenticity tokens
+SESSION_DAYS_TRIM_THRESHOLD=365
 enable_esi=
 
 #
--- a/.env.example
+++ b/.env.example
@@ -80,8 +80,8 @@ sentry_dsn=
 
 # Miscellaneous (not required)
 # set to true to use <esi:include> tags in place of form authenticity tokens
-SESSION_DAYS_TRIM_THRESHOLD=365
 enable_esi=
+SESSION_DAYS_TRIM_THRESHOLD=365
 
 #
 # End of application environment variables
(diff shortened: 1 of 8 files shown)


.env.example
```diff
@@ -80,6 +80,7 @@ sentry_dsn=
 
 # Miscellaneous (not required)
 # set to true to use <esi:include> tags in place of form authenticity tokens
+SESSION_DAYS_TRIM_THRESHOLD=365
```
wioux: This line should be moved down so the comment about esi still makes sense.

wioux:
:+1: looks good

mfb:
The main point of building the action center was for folks to be logged in so that they can take action without entering their address. Since they might only take action a couple times a year, we probably want to give logged-in (non-admin) users a very long session time, like a year.

Re: the database writes for every page view, that gives me concerns about scalability if we had a mass day of action - a million folks signing a petition, SOPA/PIPA style :)  It would be nice to not write to database until visitor starts interacting with a form.

vbrown608:
@mfb I agree that the session time should be a year. Verifying that it's actually two weeks, and I'll increase it if so. This PR supports a session time of a year.

And yes, I was surprised when I learned that we write to the database on each page view. So far it hasn't caused problems but I'm open to not tracking views, in which case we could also try to avoid storing the csrf token for anonymous users in the backend session record. Seems like a separate question though.

wioux:
I agree we need to follow up (here and elsewhere) with some scalability improvements but I think that should be a separate PR. We should release this now to cap the related vulnerability reports.

Issue #317, opened 2017-09-05
return a more helpful error when trying to remove admin from an account that doesn't exist


lenazun:
ref: https://redmine.eff.org/issues/14783
when trying to remove admin status from an account that doesn't exist, we return an unhelpful error. perhaps we could return a nicer one? I'm also changing the documentation to avoid this case.

vbrown608:
Adding a bit more detail from the redmine ticket in case a volunteer would like to jump in:

When someone runs `rake users:remove_admin[me@example.org]` and me@example.com is not a recognized user, rails throws the following error: 

```
rake aborted!
NoMethodError: undefined method `admin=' for nil:NilClass
/opt/actioncenter/lib/tasks/users.rake:12:in `block (2 levels) in <top (required)>'
/usr/local/bundle/gems/rake-11.3.0/exe/rake:27:in `<top (required)>'
Tasks: TOP => users:remove_admin
(See full trace by running task with --trace)
```

This should fail more gracefully.

k-stewart:
https://github.com/EFForg/action-center-platform/pull/355

k-stewart:
@lenazun See what you think of the fix.

lenazun:
Looks great, thank you!

PR #355, opened 2018-02-01
Return a friendly error when admin tasks can't find user

k-stewart:
https://github.com/EFForg/action-center-platform/issues/317
When we run `rake users:add_admin[email]` or `rake users:remove_admin[email]`, and no user exists with the given email, the task used to throw `undefined method `admin=' for nil:NilClass`.

And rightly so!  

Instead, task, just tell us that you can't find the user and chill.

### DIFF
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -20,7 +20,7 @@ namespace :users do
   task :add_admin, [:email] => :environment do |t, args|
     email = args[:email]
     u = User.find_by_email(email)
-    next(puts "I couldn't find a user with the email '#{email}'.") unless u
+    abort("I couldn't find a user with the email '#{email}'.") unless u
 
     u.admin = true
     if u.save
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -6,16 +6,22 @@ namespace :users do
     end.sort
     puts admins.empty? ? "No Admin Users" : admins
   end
+
   desc "Remove admin status from an account, given an email"
   task :remove_admin, [:email] => :environment do |t, args|
     u = User.find_by_email(args[:email])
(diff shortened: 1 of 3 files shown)


wioux:
We should make sure the task still exits with a non-zero code in this case, so any script relying on it can detect whether or not it failed.

k-stewart:
@wioux Ready for another look!

spec/tasks/users_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+      it 'returns a friendly message' do
+        expect { run_task }.to output(
+          /I couldn't find a user with the email '#{email}'./
+        ).to_stdout
+      end
+
+      it 'returns failure code' do
+        run_task
+        expect(`echo $?`).not_to match(/0/)
```
wioux: I'm not actually sure what I was talking about with `$?` will do the right thing in this case. There's a [stackoverflow discussion](https://stackoverflow.com/a/24680428) that suggests that the thing to do is `expect{ run_task }` to raise a SystemExit with `status != 0`. Maybe try that?

lib/tasks/users.rake
```diff
@@ -6,16 +6,22 @@ namespace :users do
...
+
     u.admin = false
     u.save
   end
+
   desc "Add admin status from an account, given an email"
   task :add_admin, [:email] => :environment do |t, args|
     email = args[:email]
     u = User.find_by_email(email)
+    next(puts "I couldn't find a user with the email '#{email}'.") unless u
```
wioux: `abort` should be used here as well since it is an error message/condition.

k-stewart:
Future explorers: when writing tests on code that calls `abort` or `exit`, you must assert `raise_error` in order to get an accurate picture of the output.  https://github.com/rspec/rspec-core/issues/2311

Issue #280, opened 2017-01-27
Nullify reset token when password is changed


wioux:

k-stewart:
It looks like this is already the case:
```
$ user.reset_password_token
=> "397b5928ebc23f8f7be73286aa28e3abebf9f45c11930be8a47beced9af816d1"
$ user.update(password: '6uhip **Iuinwoekjdfnvkl UII')
   (0.6ms)  BEGIN
  SQL (7.1ms)  UPDATE "users" SET "encrypted_password" = $1, "updated_at" = $2, "reset_password_token" = $3, "reset_password_sent_at" = $4 WHERE "users"."id" = $5  [["encrypted_password", "$2a$10$u5N19vw/NU5mc26BAKORZ.BVP1BpU4HJC7GXjYi4lqSP96F5RyFzO"], ["updated_at", "2018-02-01 18:57:42.378405"], ["reset_password_token", nil], ["reset_password_sent_at", nil], ["id", 21609]]
=> true
$ user.reset_password_token
=> nil
```

wioux:
So it is! I can't remember why I opened this... Thanks @k-stewart.

PR #356, opened 2018-02-01
Add tests to document the ways password reset tokens can be nullified

k-stewart:
https://github.com/EFForg/action-center-platform/issues/280


wioux:
:+1:

Issue #316, opened 2017-09-05
Tooltip marker wraps at some widths


vbrown608:
![wrapping-qmark](https://user-images.githubusercontent.com/1887071/30070902-bffb7a9e-9219-11e7-82d0-f45bd6493de0.png)

k-stewart:
@vbrown608 fixed, i think.

PR #358, opened 2018-02-01
Make h3s with tooltips look better on small screens

k-stewart:
https://github.com/EFForg/action-center-platform/issues/316

Before, when the text of the h3 filled most of the container, the tooltip-triggering question mark was forced down onto a new line.  It looked funny.  Now, the text of the h3 wraps around the question mark properly.

Text change is for illustrative purposes only, and not part of the commit.

screen shot 2018-02-01 at 2 53 23 pmscreen shot 2018-02-01 at 2 53 32 pm


hughillustration:
Awesome! But I think we should keep this language.

H.

wioux:
:+1:

PR #362, opened 2018-02-05
Clean up JS templates pt. 2

k-stewart:
https://github.com/EFForg/action-center-platform/issues/175

Move move phone_field.jst.ejs into the js function that calls it


wioux:
I should have caught this in part 1, but there's some more dead code in congress-forms.js we can remove now too:

https://github.com/EFForg/action-center-platform/blob/34d38f9ee6cd3120df2e7a36389b2704230db680/vendor/assets/javascripts/congress-forms.js#L357-L361

k-stewart:
@wioux I removed that function and all references to it, and also the css for the now-removed recaptcha template.  Was that right to do?  It seemed ok, but it was also pretty complicated, and I am uncertain.

wioux:
Yep, this is good! The captcha / recaptcha cases are never activated any more it all looks fine.

Issue #161, opened 2016-08-17
Allow custom friendly_ids for actions


vbrown608:
It would be great if developers could create a custom friendly id for actions in the action center by editing the action's slug in the rails console.

Currently, that gets overwritten on save. A solution could be only overwrite if that action's title changes.

vbrown608:
This was the problem that led me to revert e633845 (e-mail from Elliot):

"Something weird seems to be going on with Action Center URLs. They're
reverting to the numerical URL when you save a change to them.

https://act.eff.org/action/reclaim-invention
is forwarding to
https://act.eff.org/action/198

I made an edit to an old net neutrality campaign and the URL changed to:
https://act.eff.org/action/194"

vbrown608:
When I set the slug and save in the Rails console, and then edit the page description and save again, the slug seems to reset. Are you seeing that as well?

PR #364, opened 2018-02-06
Allow developers to edit page slugs

k-stewart:
https://github.com/EFForg/action-center-platform/issues/161

This change was made once before, but had to be reverted because it caused some weird id bugs.  This change covers some extra cases, and hopefully won't cause the same bugs; i was unable to replicate them using what i understood of the steps, so, high hopes.

If the bugs reoccur, a next step might be to add a `custom_slug` field to the ActionPage model, and implement `should_generate_new_friendly_id?` as `!custom_slug?`.

### DIFF
--- a/app/controllers/admin/action_pages_controller.rb
+++ b/app/controllers/admin/action_pages_controller.rb
@@ -70,9 +70,6 @@ class Admin::ActionPagesController < Admin::ApplicationController
 
     add_twitter_targets
 
-    @actionPage.slug = nil
-    @actionPage.save!
-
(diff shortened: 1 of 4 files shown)


vbrown608:
(Oops, I accidentally put this on the ticket instead of the PR. Moving it here:)

When I set the slug and save in the Rails console, and then edit the page description and save again, the slug seems to reset. Are you seeing that as well?

k-stewart:
: o I'll check it out!

vbrown608:
Works great, and yay for removing a weird thing from the controller to boot :)

PR #365, opened 2018-02-07
Make the Resend Unlock page match our other devise pages

k-stewart:
https://github.com/EFForg/action-center-platform/issues/293

Includes some of the comments from https://github.com/EFForg/action-center-platform/pull/294, but not the ones that made this page look unlike the other devise pages.

screen shot 2018-02-07 at 1 35 36 pm

### DIFF
--- a/app/assets/stylesheets/users.css.scss
+++ b/app/assets/stylesheets/users.css.scss
@@ -151,10 +151,17 @@ span.green-rect {
   }
 }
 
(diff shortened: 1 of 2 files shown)


wioux:
Looks good. The only thing is we should be consistent about whether or not hash rocket syntax is used, e.g. here:

https://github.com/EFForg/action-center-platform/blob/0cef798fa712a71e5de33b0cc59bde8e598ba00e/app/views/devise/unlocks/new.html.erb#L12

PR #368, opened 2018-02-14
Lint with Rubocop

k-stewart:
* Install it
* Abide by Rubocop-GitHub's conventions
* Override those conventions where necessary or to closer match SEC
* Add it to Travis
* Add it to the rake task
* Update the Readme


k-stewart:
I still need to hook Rubocop into Travis and hopefully add a precommit hook, but that can happen in another PR

README.md
```diff
@@ -117,6 +117,62 @@ To run the full test suite, simply run `rake` with no arguments.
 
 Rspec tests are used for unit testing the app, and some integration testing. Cucumber tests are used for testing API keys, javascript tests, and feature tests.
 
+## Linting
+
+`rake` will also run our linting: currently, only Rubocop.
+
+Rubocop checks for consistent style across the Ruby areas of the codebase.  We use a modified version of [Rubocop-Github](https://github.com/github/rubocop-github) to stay consistent with [SEC](https://github.com/EFForg/sec).
+
+I like to add Rubocop to a pre-commit hook, so that I can't commit non-compliant code. Here's a nice one:
```
wioux: I don't think we should suggest this. It's a personal style whether or not every individual commit has to pass lint vs. the branch ultimately passing lint before it's merged. If we want to say anything about it then I think something like "please ensure your contributions lint correctly before opening a pull request" would be more appropriate.

k-stewart:
I did some QA and everything looked good.

wioux:
Great! The only issue I can see is that the `CsrfProtectionController` actions need to set the `@token` variable so that the new views can pick it up.

wioux:
:+1:

PR #370, opened 2018-02-15
Fix invalid file names in the photo uploader

k-stewart:
https://github.com/EFForg/action-center-platform/issues/257

Replace spaces with underscores.


vbrown608:
:100: I like this solution

PR #372, opened 2018-02-15
Rate limit the petitions form

k-stewart:
https://github.com/EFForg/action-center-platform/issues/353


vbrown608:
In production, Action Center traffic comes in through Fastly. That means that from the perspective of Rails, every request has the same IP.

In production we can use the `HTTP_X_FORWARDED_FOR` header to get the IP of the original request. In development that header may not be set, so we can check the normal IP if it doesn't exist. See lines 4-7 for an example.

k-stewart:
Thanks Vivian!

vbrown608:
Great!

PR #373, opened 2018-02-15
Fix pending Institution#import specs

k-stewart:
* Check that the action creates a delayed job
* Check that the job, once run, calls `Institution.import`
* Move assertion of the behavior of `Institution.import` into the model tests

Questions:
* In the import action, if some rows are valid and some rows are not, should we import the valid rows?
* In the case of a successful import, should we flash success?
* Currently, `Institution.import` creates duplicate `ActionInstitution`s. Seems like it shouldn't, right?
* Should Institution <-> ActionPage be a `has_and_belongs_to_many`?

### DIFF
--- a/app/controllers/admin/institutions_controller.rb
+++ b/app/controllers/admin/institutions_controller.rb
@@ -40,7 +40,7 @@ class Admin::InstitutionsController < Admin::ApplicationController
       names << params["name"]
     end
 
-    Institution.delay.import(names, @actionPage)
+    Institution.delay.import(names, @actionPage) if names.present?
 
     redirect_to [:admin, @actionPage, Institution]
   end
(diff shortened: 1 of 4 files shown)


wioux:
This looks good to me, just needs the merge conflict fixed.

1. I think validating the csv format before we process it is the right thing.
2. We could say something about how the institutions are being imported.
3. Yeah it shouldn't create multiple joins to the same action page. It looks like this only happened once, for ActionPage#198, which was subsequently deleted, so there's no issue at the moment. Though that also suggests we need `dependent: destroy` on the `ActionPage#action_institutions` association. This can all go into a separate issue.
4. I tend to avoid HABTM in general. In this case it looks like we don't even need institutions to know about their action pages, so the improvement IMO would be to remove those associations altogether.

PR #376, opened 2018-02-16
Don't try to send mail in Development

k-stewart:


wioux:
I like sending mail from development. Maybe we can check whether or not the SMTP variables were provided, and `perform_deliveries = false` only if they were not?

k-stewart:
All my SMTP variables are present, but Socket is still flipping out.  Closing this until it bothers people other than me.

PR #379, opened 2018-02-21
Add Sass Lint to README, Rake task, and Travis

k-stewart:


lib/tasks/test.rake
```diff
@@ -0,0 +1,4 @@
+desc "Run all tests (rspec, rubocop, sass-lint)"
+task :sass_lint do
+  system("yarn && node_modules/.bin/sass-lint -vq")
```
wioux: Why do we need to yarn first?

k-stewart: because ./node_modules is not tracked in git, and so every environment must run `yarn` to get it.  Unless I'm wrong!

lib/tasks/test.rake
```diff
@@ -0,0 +1,5 @@
+desc "Run sass-lint"
+task :sass_lint do
+  system("node_modules/.bin/sass-lint -vq")
+  puts "No sass linting errors!  Woo!"
```
wioux: This message is emitted regardless of the exit status of sass-lint. You need something like

    system("...") && puts("...")

or

    puts "..." if $?.success?

Issue #380, opened 2018-02-22
Cache puffing-billy requests


vbrown608:
We use the puffing-billy gem to stub requests to third-party services that are made on the front end by Javascript rather than on the backend by Rails. Currently, we manually stub requests as in `features/step_definitions/congress_steps.rb`. 

Another way to use puffy-billy is to record real requests on the first pass and save them as yml files as described here: https://github.com/oesmith/puffing-billy#caching. The library will replay them on subsequent runs. I think this would give us a more accurate picture of our services with less work on our part. I started setting it up on a branch but it still needs some work.

vbrown608:
Once this is done we can improve the feature test for congress message actions targeting specific bioguide ids by checking that an appropriate form is generated based on the response form phantomdc.

PR #381, opened 2018-02-22
Accept bioguide ids for embedded congress message actions

vbrown608:
This is for Check Your Reps - users already enter their address on that site and we don't want them to enter it second time when they complete the embedded action.

The new integration test doesn't do much right now - the stubbed phantomdc always returns the same form fields. It will get better if we do #380.


app/assets/javascripts/application/tools/congress_message.js
```diff
@@ -257,12 +257,12 @@ $(document).on("ready", function() {
...
     var emailValues = tool.data("email-values");
 
-    if (tool.data("target-bioguide-id")) {
+    if (tool.data("bioguide-ids").length > 0) {
       emailValues["$MESSAGE"] = form.find(".campaign-message").val();
 
       $("form.congress-message-rep-lookup").html("");
 
-      var bioguide_ids = tool.data("bioguide-id").split(",");
+      var bioguide_ids = tool.data("bioguide-ids").split(",");
```
wioux: The pattern here needs to include possible spaces around the comma since that's how `CongressMessageHelper` joins the array. Otherwise only the first bioguide_id is located by phantomdc since the others are prefixed by a space. The error passes very silently but you can test by making sure that multiple `fieldset.congressForms-legislator-fields` appear when multiple bioguide ids are given.

vbrown608: Thanks for catching this! Looks like it's an issue for actions that are configured to always use specific bioguide IDs as well.

I can't find a ways to pass puffy billy a post body in the stubbed phantomdc, so the integration test won't catch it (it always returns the same response for the form fields lookup endpoint). With caching enabled, puffing billy saves a different response for different post bodies, so we can check for that once #380 is done. I'll make a note in that ticket.

wioux:
The js runtime for tests doesn't understand the arrow operator :disappointed:. Otherwise this all looks good!

vbrown608:
Please hold off on merging, checking integration with Check Your Reps.

vbrown608:
Okay all set.

PR #394, opened 2018-03-03
Disable csrf protection on a few actions

wioux:
Session state, which is now stored on the back end, under some circumstances is getting shared between users due to caching. I'm temporarily disabling some controller features until I can push a branch that allows us to switch away from request caching with Fastly.


mfb:
FYI there is something called login CSRF attack, where the victim is logged in to the attacker's account. Then they end up adding their PII to the attacker's account when performing some action.

wioux:
@mfb This is temporary, CSRF protection will be restored! I'm going to submit a fix for the root issue soon.

PR #395, opened 2018-03-03
Use cookie_store for sessions

wioux:
Session state, which is now stored on the back end, under some circumstances is getting shared between users due to caching. I'm temporarily reverting session storage to use cookies until I can push a branch that allows us to switch away from request caching with Fastly.


mfb:
The thing about act.eff.org that always seemed odd is that it doesn't add Cookie to the Vary header when there's a session, so then stuff with a session can get cached for everyone.

PR #396, opened 2018-03-07
Organize cache config and restore backend sessions

wioux:
* Revert #395, switching back to stateful sessions
  * Disable forgery protection on `SubscriptionsController` and `ToolsController`. These actions are hit from forms served through our HTTP cache and won't have the correct correspondence between session and authenticity_token.
  * Clean up a few orphan routes.
  * Removes everything related to ESI, it's not viable. On the plus side we can gzip everything!
  * Improve our cache config and organize/track it better:
    * Update VCL config to remove Cookie/Set-Cookie headers from requests except where necessary. Otherwise there'll be no caching!
    * Move all cache header config into a yaml file config/response-headers.yml. This should make it easier to tell what pages are/aren't being cached, for how long, etc., rather than having to both grep through the app and check the Fastly UI.

This branch/vcl config is deployed to staging for testing.


vbrown608:
I like this solution. We should create a wiki page documenting the fastly/varnish situation (and the reason for disabling CSRF of anonymous forms) and link to it from the varnish config and maybe also the Readme.

Other than that, looks good to me!

wioux:
Ok done, thanks. I do want to circle back to the CSRF issue soon. We can at least check that the Origin header is as expected.

vbrown608:
:scroll: :+1:

PR #400, opened 2018-03-13
Don't pass referrer header to twitter on action share link

vbrown608:
The Action Center embed url can be super huge. In general, we should pass a link to a stylesheet instead of putting all the styles in the url, but this will help prevent HTTP 431 from Twitter when the url gets big.

This PR also dried up some views. Call actions previously didn't have a donate link, while all the other actions did. This adds a donate link to call actions.

### DIFF
--- a/app/views/tools/_call.html.erb
+++ b/app/views/tools/_call.html.erb
@@ -102,28 +102,7 @@
             </div>
           </div>
 
-          <div class="vertical-share">
-            <p>Once you're done, help spread the word:</p>
-            <a href="<%= twitter_share_url(@actionPage) -%>" class="twitter twitter-button">
-              <div class="sicon" id="stwit"></div>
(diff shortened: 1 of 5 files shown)


wioux:
The referer header is still sent, since the click on that link is caught and handled with javascript:

https://github.com/EFForg/action-center-platform/blob/58efcede657445f84a95fd2311f5c2306b51ce71/app/assets/javascripts/application/social-buttons.js#L14-L18

I think we should prioritize deprecating embedded css rather than try to fix this one.

:+1: the view cleanup

vbrown608:
Ah, good point. I'll revert the commit that sets noreferrer. 

I tried to pass a standalone stylesheet to Stop Sesta last night, but ran into CSP issues in production. I'll see if I can resolve that.

PR #401, opened 2018-03-13
When we can't do verify_authenticity_token, verify origin header instead

wioux:
As recommended in the [CSRF cheat sheet](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#Verifying_Same_Origin_with_Standard_Headers). Follow up to #396.


wioux:
Thanks, I added a few tests.

PR #403, opened 2018-03-14
Serve S3 files over HTTPS

### LINKED ISSUE #385: Some static assets loaded over http
We should be loading everything from s3 over https to avoid mixed-content warnings

vbrown608:
The tiniest PR :)

Fixes #385


maximillianh:
it's so smo

PR #406, opened 2018-03-26
Move congress_forms.js to /app/assets

### LINKED ISSUE #290: congress-forms.js has diverged from source
https://github.com/EFForg/congress-forms.js/

vbrown608:
This resolves #290 somewhat. It moves congress_forms.js from vendor to app - it's diverged from source, and I think we should just consider it to be part of the Action Center now. I will mark the old upstream repo as deprecated - this is the only place where it's used. 

I also did some light cleanup, mainly
- Pull out captcha logic, since we don't support captcha anymore
- Remove debug blocks in a favor or a cucumber integration test

There's still not a great logical separation between the contents of congress_message.js and congress_forms.js. I'd like to move some of this to the backend, but I'm opening this for now to avoid dropping a giant PR.


app/assets/javascripts/application/tools/congress_message.js
```diff
@@ -162,22 +161,6 @@ $(document).on("ready", function() {
...
           $("input[type=submit]").slideUp(400, height_changed);
-          if ($('[id="$CAPTCHA_SOLUTION"]', legislatorFieldSet).length === 0) {
-            setTimeout(function () {
-              emailsSent++;
-              if (emailsSent === rep_ids.length) {
-                allEmailSent();
-              }
-              $("#congressForms-common-fields .form-group").slideUp(400, height_changed);
-              $(".form-group", legislatorFieldSet).slideUp(400, height_changed);
-              legislatorFieldSet.find("span.info-circle").replaceWith('<span class="info-circle success">Sent <i class="icon-ok-circle"></i></span>');
```
wioux: Without this the "Submitting \<spinning activity circle\>" indicator never resolves. The ideal place to move this would be into the `onLegislatorSuccess` callback, but I'm guessing there's a reason it was made to update the UI unconditionally on a short timeout here, probably because congressforms can sometimes take quite a while to complete a request (and in any case the message might get queued and so `onLegislatorSuccess` won't get fired).

features/step_definitions/action_page_steps.rb
```diff
@@ -4,18 +4,41 @@
...
-When(/^I fill out the required fields for Nancy Pelosi$/) do
+When(/^I fill in the required fields for Nancy Pelosi$/) do
   select("Ms.", from: "$NAME_PREFIX")
   fill_in("$NAME_FIRST", with: "Buffy")
   fill_in("$NAME_LAST", with: "Summers")
   fill_in("$EMAIL", with: "bsummers@ucsunnydale.edu")
   select("Congress", from: "$TOPIC")
 end
 
+When(/^I fill in the required fields for Nancy Pelosi and Kamala Harris$/) do
```
wioux: A few tests are failing because they use "I fill *out* the required fields..." for these steps.

wioux:
We have to restore the UI behavior I mentioned but otherwise I think this is a good move. :+1: getting rid of the inline debug logic.

PR #407, opened 2018-04-04
Migrate to aws-sdk v2

k-stewart:
I believe that this PR represents all necessary changes to upgrade AWS-SDK to v2.  However, I can't upload files from my local environment, so I can't test that it works.  Maybe my secrets are wrong?  

Anyway, the action returns success and does not return "Attempt to get S3 object failed.", so I'm somewhat confident.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -72,11 +72,14 @@ GEM
       builder
       mime-types
       xml-simple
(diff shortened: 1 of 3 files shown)


k-stewart:
whoops, does not work.

Issue #374, opened 2018-02-16
Inconsistent tooltip icons 


k-stewart:
![screen shot 2018-02-16 at 11 02 39 am](https://user-images.githubusercontent.com/1065956/36324313-eef4527e-1308-11e8-9691-d7e0593e153f.png)

vbrown608:
Closed by #410

PR #410, opened 2018-04-13
make action tooltips consistent - issue #374

fv42wid:
Hi there - I'd like to get involved with this organization and this issue looked like a good place to get started.  Let me know if anything does not comply with your workflow.


vbrown608:
Thanks for taking a look at this! I agree that the privacy notice popover needs to have its height and width bumped up to 20px. I think we could keep the original markup and fix the height and width for the `privacy-notice-popover` class without running into problems elsewhere.

fv42wid:
Thank you for the feedback.  I added a commit to update the original class instead of adding a new class.

vbrown608:
Looks good!

PR #416, opened 2018-05-17
GDPR compliance

### LINKED ISSUE #413: Update subscription checkboxes for GDPR compliance
On the registration form at https://act.eff.org/register the default state of both radio buttons should be 'false' for GDPR compliance purposes.

k-stewart:
Closes #413. 
screen shot 2018-05-16 at 5 34 27 pm
screen shot 2018-05-16 at 5 34 38 pm

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -57,6 +57,8 @@ class User < ActiveRecord::Base
   end
 
   def percentile_rank
+    return 0 unless record_activity?
+
     user_action_counts = Rails.cache.fetch("user_action_counts", expires_in: 24.hours) {
       User.select("users.id, count(ahoy_events.id) AS events_count")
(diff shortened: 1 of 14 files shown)


app/models/user.rb
```diff
@@ -57,6 +57,8 @@ def name
   end
 
   def percentile_rank
+    return 0 unless record_activity?
```
vbrown608: I think it might be preferable to hide the stats widget on `/account`, rather than display all zeros. Seeing zeros there might confuse users, and if they took actions before they opted out of tracking they'll see nonzero, but still incorrect, numbers.

vbrown608:
This looks good to me. @k-stewart let me know if you'd to update based on the comment above now or merge. I feel somewhat strongly about that stats box (but down to discuss!) but we could always do a follow-up and deploy this in the name of GDPR.

k-stewart:
@vbrown608 ready for another look.

PR #419, opened 2018-06-25
Fix password reset

k-stewart:
https://redmine.eff.org/issues/17730
Shahid couldn't reset his password; when he tried, it raised a 500.  I got the same result.  Steps to replicate:

1. go to https://act.eff.org/login
2. click 'forgot your password'
3. enter your email and click 'send me reset password instructions'
4. receive email; click link. This takes you to a page like "https://act.eff.org/password/edit?reset_password_token=something"
5.enter your new password twice
6. click "change my password"
7. oh noes: "500 Internal Server Error If you are the administrator of this website, then please read this web application's log file and/or the web server's log file to find out what went wrong."

When I looked at the logs, I saw that PUT /passwords was returning 422 with`Can't verify CSRF token authenticity`.  The params did include AuthenticityToken.  Stack Overflow suggested that it might be a cached token.  I looked into how we cache with Fastly, and turned off caching for password reset endpoints.

I'm not sure this will work!  I'll talk to Vivian about it.

### DIFF
--- a/config/response-headers.yml
+++ b/config/response-headers.yml
@@ -33,6 +33,8 @@ devise/confirmations:
 
 devise/passwords:
   new: *dont_cache
+  edit: *dont_cache
+  update: *dont_cache
 
 devise/unlock:
   new: *dont_cache


vbrown608:
It does sound like this could be a cacheing issue. We shouldn't be cacheing any pages for logged in users - that's set here: https://github.com/EFForg/action-center-platform/blob/master/varnish.vcl#L10-L13.  That's why `edit` and `update` shouldn't need to be explicitly setup to not be cached.

k-stewart:
Hmm, but those pages are accessible to - in some cases, explicitly! - users who are not logged in.

vbrown608:
Oh ahh right of course! I blame vacation brain. This seems right to me in that case. Nice job sleuthing through the fastly docs/setup.

k-stewart:
Ah, cool.  Ty @wioux. Closing this PR in favor of one that does that.

vbrown608:
@wioux could you say more about what that block is doing? I'd like to be able to debug cacheing issues like this in the future.

k-stewart:
Ah, cool.  Ty @wioux. Closing this PR in favor of one that does that.

wioux:
The default VCL config has rules to bypass the cache for requests with a `Cookie` header, and not to cache responses that include a `Set-Cookie` header. Because Rails sets cookies on every page, we have custom VCL to delete cookie headers before those rules are encountered, except for actions that really need them (forms with strong CSRF protection, and the ahoy analytics routes).

When we don't make that exception (like with this bug), we send the client a page with a csrf token, but delete the cookie containing the associated session id. The client doesn't get the cookie, and when it submits the form, it sends the csrf token but there is no session to match it against, which causes the error.

There's some more info about varnish and cookies at [1] and [2].

I put the whitelisted routes in VCL but in retrospect there are better ways to do it. We could do something like this instead

```vcl
if (beresp.http.Surrogate-Preserve-Cookie != "1") { 
     unset beresp.http.Set-Cookie; 
}
```

and move the responsibility of knowing which action absolutely require cookies to the Rails app.

[1] https://varnish-cache.org/docs/2.1/tutorial/cookies.html
[2] https://www.fastly.com/blog/common-causes-poor-cache-hit-ratio-and-how-deal-them

vbrown608:
Thanks for the links! Very helpful.

PR #421, opened 2018-06-28
Change logout from a GET to a DELETE request

esoterik:
This is for redmine #17076

### DIFF
--- a/app/assets/stylesheets/application/_base.css.scss
+++ b/app/assets/stylesheets/application/_base.css.scss
@@ -76,6 +76,24 @@ a:hover {
   }
 }
 
+.nav-form-link {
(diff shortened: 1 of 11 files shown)


vbrown608:
Looks like signing out by DELETE is the default and we're overriding it here: https://github.com/EFForg/action-center-platform/blob/master/config/initializers/devise.rb#L228

I suspect that this change was made to allow logout via html links for no-js users. When I disable Javascript and click logout I see `No route matches [GET] "/logout"`

Rails is adding `data-method=delete` to the link and submitting it by ajax, but I don't think vanilla html links support any method besides GET.

vbrown608:
Great! Nice job with the styles.

PR #448, opened 2019-01-12
Fix Docker build

vbrown608:


mfb:
What I have done on some other projects, to get debian-slim to work, is create the expected man directory myself in the Dockerfile.  Kind of silly but it works.

PR #449, opened 2019-01-14
Mailing list honeypot captcha

vbrown608:
* Protect mailing list signup endpoints with invisible captcha
* Surface errors from mailing list signup, so we can disable signups if necessary

I didn't put invisible captcha on congress or call actions - I think any bot submissions would have to be tailored to those forms, in which case the invisible captcha wouldn't help.

### DIFF
--- a/app/assets/javascripts/application/welcome.js
+++ b/app/assets/javascripts/application/welcome.js
@@ -47,12 +47,14 @@ $(document).ready(function() {
     },
     complete: function(xhr, data, status) {
       $('.progress-striped').hide();
+      var message;
       if (status === "success")
-        return $(App.Strings.successfulEmailSubmission).insertAfter(this);
+        message = App.Strings.successfulEmailSubmission;
       else if (data.responseJSON.message)
-        return $(data.responseJSON.message).insertAfter(this);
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/application/welcome.js
```diff
@@ -47,10 +47,12 @@ $(document).ready(function() {
     },
     complete: function(xhr, data, status) {
       $('.progress-striped').hide();
-      if(status === "success")
+      if (status === "success")
         return $(App.Strings.successfulEmailSubmission).insertAfter(this);
+      else if (data.responseJSON.message)
+        return $(data.responseJSON.message).insertAfter(this);
```
wioux: I think the message here needs to get wrapped in a tag? So that the jquery helper makes a new node rather than interpreting it as a selector.

mfb:
If you have timestamp checking via javascript, then you can still cache the page.

vbrown608:
Oh good point @wioux! I think let's start with just the honeypot and see if the logging @mfb added on the civi end of things gives us any insight into whether the action center is causing problems. I'm also talking to activism about adding back double opt-in.

PR #450, opened 2019-01-15
Fix opt-in parameter name for newsletter signups

wioux:


wioux:
It's possible but would take an update to ed/kittens as well, as there is special handling of contacts coming from action center. Want to open an issue there?

PR #469, opened 2019-02-15
Normalize date ranges for analytics charts

### LINKED ISSUE #465: Normalize date ranges for analytics charts
The action page analytics charts begin with the first date in the requested range with any events. We should begin with the first date/time in the requested range, regardless of event count, instead so that it's easier to compare across charts.

wioux:
Fixes #465, normalizing the analytics charts so they always show the complete range.

### DIFF
--- a/app/models/ahoy/event.rb
+++ b/app/models/ahoy/event.rb
@@ -20,17 +20,16 @@ module Ahoy
     after_create :record_civicrm
 
     def self.group_in_range(start_date, end_date)
-      zone = Time.zone.now.zone
       if (end_date - start_date) <= 5.days
         group_by_hour(
           :time,
-          format: "%Y-%m-%d %H:00:00 #{zone}",
+          format: "%Y-%m-%d %H:00:00 %Z",
           range: start_date..end_date.tomorrow
         ).count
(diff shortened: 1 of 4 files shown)


vbrown608:
Thanks for working on this. It's handy to get to test it in a different timezone from you!

I really like the idea of serving SF time to everyone. Sadly, chartkick doesn't seem to respect the timezone in the JSON. I see `"2019-03-10 00:00:00 PDT":0` in the response, but it still renders using my OS's timezone:
zonestart

It seems like we can force it to use the JSON more strictly by passing `discrete: true` to the `line_chart` helper (though the scale marks get less pretty - maybe there's a way to transform them). Or maybe we could fudge it on the backend by not including a timezone in the output at all?

wioux:
Thanks for the feedback! Good idea, I agree that's the right behavior. I went ahead with `discrete: true` and changes to the label format. I'd have rather let the front end decide how the labels are formatted, but it looks like this is the only way. The library converts back and forth between `Date`s and unix time internally, losing the original timezone and always ending up in the browser's (and initial values with no timezone are treated as UTC).

Issue #475, opened 2019-03-19
Migrate away from rails-assets


esoterik:
rails-assets isn't super well supported anymore, and I ran into [this issue](https://github.com/tenex/rails-assets/issues/238) while working on the rails 5 upgrade. It'd be good to switch to npm for managing JS packages and probably the `bootstrap-sass` gem for bootstrap.

PR #485, opened 2019-05-01
[WIP] Rails 5!

esoterik:
(WIP until QA)

Major changes include: 
* Removing the s3_cors_fileupload gem and just using aws-sdk instead. This also meant getting the jquery file upload JS from a separate source; this PR includes a very minimal npm setup in order to keep using the file upload plugin we were using previously. 
  - There's already an issue to migrate away from rails-assets for JS dependencies (#475), as rails-assets is no longer well-maintained.
* Switching to the bootstrap-sass gem instead of the rails-assets version of bootstrap

Will open issues for:
* Restoring sortable tables (used to reorder topics) -- the jquery plugin we were using here hasn't been maintained and requires a JQuery version with many vulnerabilities. 
* Deprecations that will cause things to break in Rails 5.1
* Deprecations that will cause things to break in Rails 5.2


config/environments/production.rb
```diff
@@ -77,6 +77,7 @@
   #config.assets.precompile += %w( .svg .eot .woff .ttf )
   config.assets.precompile = config.assets.precompile.map!{ |path| path == /(?:\/|\\|\A)application\.(css|js)$/ ? /(?<!jquery-sortable)(?:\/|\\|\A)application\.(css|js)$/ : path }
   config.assets.precompile += %w( admin/_bootstrap-theme.min.css admin/admin.css admin.js admin-head.js )
+  config.assets.precompile += %w( congress-images-102x125/i/P000197.jpg )
```
wioux: Is this needed?

esoterik: I think so? The tweet action feature was failing in CI with an "asset not declared to be precompiled in production" error -- see [this build](https://travis-ci.org/EFForg/action-center-platform/builds/526691906#L3506)

wioux: Ah ok! I was unsure since it's the image for a specific rep (Pelosi). I've been testing and it looks like we'll need to precompile the whole set:

```ruby
  config.assets.precompile += %w( congress-images-102x125/i/* )
```

You can test on /action/tell-congress-to-investigate-cbp-social-media-monitoring -- when you enter a non-SF address, the image request 500s.

wioux:
Tests were initially failing for me due to missing precompiled assets. I got them passing by moving the assets config (currently duplicated in `config/environments/{development,production}.rb`) into an initializer.

vbrown608:
I'm able to successfully add photos and select them from the gallery now, but the featured photo for the action doesn't seem to save when I save the action page.

wioux:
:+1: looks good to me

PR #507, opened 2019-05-29
Search/filtering for action pages in admin area

wioux:
This PR adds a form to the `admin/action_pages#index` page which filters the list by search text or action type.


app/controllers/admin/users_controller.rb
```diff
@@ -28,8 +28,7 @@ def update
 
   def filtered_users
     if params[:query].present?
-      User.where("LOWER(email) LIKE %?% OR LOWER(first_name || ' ' || last_name) LIKE %?%",
-                 params[:query], params[:query])
+      User.search(params[:query])
```
vbrown608: \*chef's kiss\*

app/models/action_page.rb
```diff
@@ -36,6 +54,24 @@ class ActionPage < ActiveRecord::Base
 
   scope :categorized, ->(category) { joins(:category).where(categories: { title: category }) }
 
+  def self.type(*types)
```
vbrown608: This looks like it was written to support multiple types, but I'm only able to filter on a single type because the form utilizes a radio button element. Is that intentional?

app/models/action_page.rb
```diff
@@ -36,6 +54,24 @@ class ActionPage < ActiveRecord::Base
...
+    scopes = Array(types).flatten.map do |t|
+      unless %w(call congress_message email petition tweet redirect).include?(t)
+        raise ArgumentError, "unrecognized type #{t}"
+      end
+
+      where(:"enable_#{t}" => true)
+    end
+
+    seed = scopes.shift || all
```
vbrown608: It looks like `inject`/`reduce` can be [called without a seed](https://apidock.com/ruby/Enumerable/inject) - could 66-74 be replaced with
```
    return all if types.empty?
    scopes.inject{ |scope, cond| scope.or(cond) }
```
?

app/views/admin/action_pages/_panel_action_pages.html.erb
```diff
@@ -1,4 +1,59 @@
...
+
+      <%= radio_button_tag "action_type", "tweet", params[:action_type] == "tweet" %>
+      <%= label_tag("action_type_tweet") do %>
+        <i class="icon on ion-social-twitter"></i>
+        <i class="icon off ion-social-twitter-outline"></i> Tweets
+      <% end %>
+    </fieldset>
+
+    <div class="query">
+      <%= text_field_tag :q, params[:q], placeholder: "Filter by title, description, action text, etc.", autocomplete: "off", class: "form-control input-sm" %>
```
vbrown608: We should label the "filter" field using one of these approaches: https://www.w3.org/WAI/tutorials/forms/labels/#hiding-label-text.

app/models/action_page.rb
wioux: Ooh cool. `scopes.inject(:or)` also works :D

app/models/action_page.rb
wioux: Sort of. I was experimenting with checkbox vs. radio behavior for the action type filter. The UX felt better with radios, but I kept this method as it was when it supported both.

vbrown608: Cool, that sounds fine to me. I thought the radio worked well from a UX perspective too.

wioux:
Thanks for the feedbacK!

wioux:
Aah whoops. Thanks for the feedback, this is ready for review again.

PR #511, opened 2019-05-30
Reset password for unconfirmed users

### LINKED ISSUE #471: Password isn't set when registering (again) for an account
When an email is already taken by an un-confirmed user, and a new registration is attempted with the same address, we [swallow the error](https://github.com/EFForg/action-center-platform/blob/120e3795f8f21116e47076e2a576e427c5801db1/app/controllers/registrations_controller.rb#L24) that would appear and mail out another confirmation link instead. When the user confirms their account, the password is still whatever was entered during the first registration attempt.

vbrown608:
When a user has an unconfirmed account and they try to register again with the same e-mail, update the password associated with the confirmed account.

Closes #471

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -24,7 +24,13 @@ class RegistrationsController < Devise::RegistrationsController
     resource.errors.delete(:email)
 
     if resource.errors.empty?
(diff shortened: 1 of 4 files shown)


wioux:
:+1: thanks for writing tests!

PR #514, opened 2019-05-30
Allow sweetalert cookie

### LINKED ISSUE #470: Registration instructions are hidden due to caching
After registering for an account, users are redirected to `/` where [instructions should pop up](https://github.com/EFForg/action-center-platform/blob/120e3795f8f21116e47076e2a576e427c5801db1/app/controllers/registrations_controller.rb#L50) telling them to check for a confirmation email. In production, they wind up hitting the fastly cache for act.eff.org, so the notification doesn't appear.

vbrown608:
Fixes #470 

Sweetalert shows an alert on redirect by putting it in a cookie on the 302 response. We need to retain the Set-Cookie header on 302 response (line 27) and retain the cookie itself on the redirected request to `/` (line 15).

This is deployed to our staging Fastly config.

### DIFF
--- a/varnish.vcl
+++ b/varnish.vcl
@@ -12,6 +12,10 @@ sub vcl_recv {
       return(pass);
   }
 
+  if (req.http.cookie ~ "^sweetAlert") {
+      return(pass);
+  }
+
   if (req.url !~ "^/(ahoy)/") {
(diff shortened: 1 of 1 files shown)


lschatzkin:
I haven't tested this but lgtm.

PR #528, opened 2019-06-07
Redesign 2019

lschatzkin:


app/assets/stylesheets/action_page.scss
```diff
@@ -1,30 +1,9 @@
...
-
-.redtop {
-  padding: 2em 0 0;
-  h1 {
-    font-size: 2.5em;
-    font-weight: normal;
-    line-height: 1.2;
-    margin-top: 0;
-  }
+* This is a manifest file that'll be compiled into action_page.css, which will include all the files
```
wioux: It's probably better to remove these comments rather than update them. I'm not sure where they came from originally, but it's not true anymore -- the only files compiled into .css are application.scss and the stylesheets listed in `Rails.application.config.assets.precompile`.

wioux:
Can we stop the new styles bleeding into the admin area? That'll have different a different look from the user facing part of the site -- not even dark themed necessarily (:crossed_fingers:), from what we saw in the mockups last week.

app/assets/stylesheets/application/_footer.scss
```diff
@@ -0,0 +1,53 @@
+.footer {
+  background: black;
+  color: $gray;
+  display: block;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  padding: 2rem 1rem 5rem 1rem;
+  text-transform: uppercase;
```
wioux: I'm not a fan of uppercase text, but if the design requires it I'd rather we attach the style to a more specific element (e.g. `.footer a` instead of the whole footer area, or `.featured-action h2` instead of all h2).

app/assets/stylesheets/application/_variables.scss
```diff
@@ -1,44 +1,47 @@
...
+
+/* colors */
+$eff-red: #ec1e1e;
+$red: #FF0000;
+$dark-red: #b80101;
+
 $blue: #0f6499;
 $green: #7dcc00;
-$sans-serif: 'Roboto';
+$off-white: #EAEAEA;
```
wioux: Would it be a lot of work at this point to add semantic names to the theme colors and use those in most instances, rather than the color name directly? Something like

```scss
// _variables.scss
$off-white: #EAEAEA;
$gray-dark: #999999; // footer text
$charcoal: #2D2D2D;

// _theme.scss
$page-background: $charcoal;
$tool-background: $gray-dark;
$text-color: $off-white;
$header-color: white;

// _base.scss
body {
  background: $page-background;
  color: $text-color;
}

// _action_page.scss
#tools .tool {
  background: $tool-background;

  .tool-title {
    color: $header-color;
  }
}
```

app/assets/stylesheets/action_page.scss
lschatzkin: agreed.

app/assets/stylesheets/application/_footer.scss
lschatzkin: yup

app/assets/stylesheets/application/_variables.scss
lschatzkin: We could. I guess I've always preferred just name the colors.  Perhaps we could name the variables by color, and then have them extended for semantic purposes? I've seen a lot of sites where a color variable will be something like "page-background" but then gets used for other purposes, and it gets weird.

wioux:
I realized after making the comment about these styles affecting the admin area -- it's probably better to fix that by starting fresh with a new stylesheet, and then not even include application.css in that section. It'll have a completely different look from the user facing part of the site anyway.

app/assets/stylesheets/application/_variables.scss
```diff
@@ -1,44 +1,47 @@
...
-$darkgrey: #353535;
-$lightgrey: #ddd;
-$mediumgrey: #636363;
+$base-font-size:   16px;
+$small-font-size:  $base-font-size * 0.875;
+$base-line-height: 1.5;
+
+/* colors */
+$eff-red: #ec1e1e;
+$red: #FF0000;
```
wioux: White against `#FF0000` is pretty harsh to read IMO. It fails WCAG AA for normal text. The shade we're currently using (`#CC0000`) is better, but contrasts poorly with `#949494` the tool background color (and fails the WCAG check for UI components). Not sure what to suggest here -- just flagging it as a drawback.

lschatzkin: Thanks, Peter, I noticed this is well and will discuss with design.

app/assets/javascripts/application/main.js
```diff
@@ -20,12 +20,3 @@
     }
   }
 }());
-
-webshims.setOptions('basePath', '/webshims/shims/')
```
esoterik: now that we're deleting this, we can also get rid of `lib/tasks/webshims_asset_compile.rake`, L78 (`RUN bundle exec rake webshims:update_public`), and the webshims-rails gem!

PR #530, opened 2019-06-07
521 redtop

lschatzkin:
- replace class "redtop", name is meaningless in new style. Removed where it seemed unnecessary.
- replace some media calls with variables. Others will have to be reexamined further in the process, as they use max-width (not mobile-first)


app/views/welcome/index.html.erb
```diff
@@ -1,12 +1,9 @@
 <% content_for :head do -%>
   <%= render 'welcome/meta_tags' -%>
 <% end -%>
-<div class="redtop home">
+<div class="home">
   <div class="container">
     <h1><%= t :front_page_header_html -%></h1>
-    <% unless user_signed_in? -%>
-      <%= render 'subscriptions/form' -%>
-    <% end -%>
```
vbrown608: Are we sure this is going away for anonymous users?

vbrown608:
So nice to see the arrival of variables for breakpoints!!

app/views/welcome/index.html.erb
lschatzkin: My bad, restoring.

esoterik:
I got an error loading the home page on this branch -- `Sass::SyntaxError - Undefined variable: "$lightgrey"`, which I think has been showing up in the CI as well. I think I found the culprit, though!

`app/assets/stylesheets/admin/_events.scss`, line 6: `border: 1px solid $lightgrey;`.

lschatzkin:
> I got an error loading the home page on this branch -- `Sass::SyntaxError - Undefined variable: "$lightgrey"`, which I think has been showing up in the CI as well. I think I found the culprit, though!
> 
> `app/assets/stylesheets/admin/_events.scss`, line 6: `border: 1px solid $lightgrey;`.

I swear I've changed that twice already and it keeps reverting... okay, let's hope third time is a charm.

esoterik:
sweet, that fixed that part of the build!

there's a new test failure, but i think it's just a matter of deleting the offending line in `action_pages/show.html.erb` -- action pages no longer have background images, correct? if that's not the case this can definitely be fixed later.

(also all the changes in this PR look good, it's just a matter of fixing the small build errors)

esoterik:
leaving the cucumber failures for now, merging!

PR #533, opened 2019-06-12
Display congress message fills in events table

vbrown608:
* Remove the old congress messages table, which loaded data on the front end from congress_forms_api
* Move the logic for make calls to congress_forms_api from view helpers to lib + the CongressMessageCampaign model
* Render congress message fills in the action page events table. For the site analytics page, I swapped in congress forms actions recorded in ahoy...it seems unlikely that this type of specificity will be necessary there.

Marking this WIP because I want to make sure it works well with the proposed change to congress_forms_api before merging.


app/controllers/admin/events_controller.rb
```diff
@@ -5,6 +5,9 @@ def index
     if params[:type].blank?
       @data = events.group_by_type_in_range(start_date, end_date)
       @columns = Ahoy::Event.action_types(action_page)
+      if action_page.present? && action_page.congress_message_campaign.present?
```
wioux: It might be better to check `enable_congress_message?` instead, since the form for action pages will create tool associations even when they're not used.

PR #534, opened 2019-06-12
Loop over events to render stats charts

vbrown608:
Just a little refactor - now that we have a list of possible events, we can iterate over them to render the charts.


app/views/admin/events/_action_page_events.html.erb
```diff
@@ -51,34 +53,6 @@
...
-        <% if @actionPage.enable_tweet? && @actionPage.tweet %>
-          <h3>Tweet Tool</h3>
-          <%= line_chart(admin_action_page_events_path(@actionPage, type: :tweets),
-                         height: "200px", discrete: true) %>
-        <% end %>
-
-        <% if @actionPage.enable_petition? && @actionPage.petition && @actionPage.petition.persisted? %>
-          <h3 class=clearfix>
-            Petition Tool
-            <% if can?(:export_signatures) %>
```
vbrown608: I've removed this permissions check. It falls through to the default check that the user is an admin, which they need to be to view this page. We're not using this permissions system extensively and the next link over lets them view the full list of signatures anyway.

app/views/admin/events/_action_page_events.html.erb
```diff
@@ -16,21 +16,23 @@
...
+          <% end %>
 
-        <% if @actionPage.enable_congress_message? && @actionPage.congress_message_campaign %>
-          <h3>Congress Messages Tool</h3>
-          <%= line_chart(admin_action_page_events_path(@actionPage, type: :congress_messages),
+          <%= line_chart(admin_action_page_events_path(@actionPage, type: type),
                          height: "200px", discrete: true) %>
+        <% end %>
 
+        <% if @actionPage.congress_message_campaign_id %>
```
wioux: Here's another spot where `enable_congress_message?` will be more accurate.

app/views/admin/events/_action_page_events.html.erb
```diff
@@ -16,21 +16,23 @@
     <hr />
     <div class="row">
       <div class="col-sm-12">
-        <h3>Action Page Views</h3>
-        <%= line_chart(admin_action_page_events_path(@actionPage, type: :views),
-                       height: "200px", discrete: true) %>
+        <% Ahoy::Event.action_types(@actionPage).each do |type| %>
+          <h3><%= humanize_event_type(type) %></h3>
```
wioux: Not a merge blocker, but `humanize_event_type` looks a lot like the built-in `t` helper. It might be more conventional here to call out to that instead. You'd just need to add something like

```ruby
  ahoy:
    views: Action Page Views
    emails: Email Actions
    congress_messages: Congress Message Actions
```

to en.yml and call `t(type, scope: :ahoy)`.

vbrown608: nice!

PR #544, opened 2019-06-19
Design cleanup

lschatzkin:
a general mid-project cleanup of the front end styling, more to come


esoterik:
this looks great!!

when I ran it locally, the `.link-div` on the homepage was the size of the whole page, so clicking anywhere took me to an action. it looks like this is the case for all of the action page links on the homepage

lschatzkin:
Good spotting! Fixed on the home page, and am not seeing css for this on the action pages

db/schema.rb
```diff
@@ -1,4 +1,3 @@
-# encoding: UTF-8
```
wioux: Can you revert all these changes? db/schema.rb shouldn't be updated in a styling update.

lschatzkin: Agh! Sorry.

wioux: We shouldn't delete this file, only undo the changes to it made in e6110236761222ec13c4cf14b5bbb94540872cbf.

lschatzkin:
I've cleanup up/restored a few things to leave bootstrap alonge as much as possible, ready for re-review

esoterik:
we lost text links to actions on the homepage at some point -- sorry for not catching it before! it's in L42 of `app/views/welcome/index.html.erb` -- right now it's just an `h3` but it should also be a link.

lschatzkin:
> we lost text links to actions on the homepage at some point -- sorry for not catching it before! it's in L42 of `app/views/welcome/index.html.erb` -- right now it's just an `h3` but it should also be a link.

Good catch! I'll fix this now.

lschatzkin:
Actually, I'm not seeing anything different b/t this branch and master...

esoterik:
oh weird, i guess on master the `span.link-div` covers the whole div and makes it clickable. I don't think we should go back to that (I don't expect whole paragraphs to be links when I'm interacting with a site), but I do expect the title of the action to be clickable, so I think we should still add that link to the `h3`

lschatzkin:
Yep, I remember the link.div weirdness now, I'll add the link to h3.

lschatzkin:
I messed around with a few other things while I was in there.

esoterik:
lgtm, merging!

PR #549, opened 2019-06-21
545 action page

lschatzkin:
Many changes to the forms. Some cleanup still needs to happen.

Test with as many actions as you can stand (tweet, email, etc.)


wioux:
The line-height of action tool titles make them jumble up when they span more than one line. We should avoid using absolute units for line-height because they tend to break like this when changing font-size (unless 

Screenshot_2019-06-24_10-26-00
you also write a new line-height for every change in font-size). We have 48 different line-height rules in the style sheet now, I think we should aim for more like 5-10.

wioux:
The tool gets pushed very far right for me and gets more cut off as the screen gets narrower.

Screenshot_2019-06-24_10-41-07

app/assets/stylesheets/action_page.scss
```diff
@@ -20,64 +36,47 @@
...
-      display: inline-block !important;
-    }
-  }
-}
-
 #tools {
-
   .tool {
-    background-color: #ffffff;
+    background-color: $gray-dark;
```
wioux: `#999999` fails WCAG guidelines when white or $off-white is in the foreground, at all text sizes. https://webaim.org/resources/contrastchecker/

We should ask design for a better color here or get creative ourselves. I think it might be hard to address accessibility concerns within the gray-on-gray-on-gray theme.

lschatzkin:
> The line-height of action tool titles make them jumble up when they span more than one line. We should avoid using absolute units for line-height because they tend to break like this when changing font-size (unless we have a new line-height for every change in font-size, but that becomes difficult to manage and is unnecessary IMO). We have 48 different line-height rules in the style sheet now, I think we should aim for more like 5-10.
> 
> Screenshot_2019-06-24_10-26-00

Thanks for noticing the plethora of line heights. I'll make mostly all of those disappear.

app/assets/stylesheets/action_page.scss
lschatzkin: I'll bring this up with Design.

lschatzkin: Found a very close color that passes (#949494), adding it.

lschatzkin:
> > The line-height of action tool titles make them jumble up when they span more than one line. We should avoid using absolute units for line-height because they tend to break like this when changing font-size (unless we have a new line-height for every change in font-size, but that becomes difficult to manage and is unnecessary IMO). We have 48 different line-height rules in the style sheet now, I think we should aim for more like 5-10.
> > Screenshot_2019-06-24_10-26-00
> 
> Thanks for noticing the plethora of line heights. I'll make mostly all of those disappear. The action pages ones I'll do here, the others up other pages.

Okay, I've made it so at least it does not jumble, I'll add another ticket to address line heights in general.

app/assets/stylesheets/action_page.scss
wioux: Thanks. I'm going to try to start a more general conversation about this with design. Overall the gray theme feels very low-contrast to me.

lschatzkin:
I've done a some cleanup on this branch, ready for re-review

esoterik:
will review after conflicts are fixed!

wioux:
The sticky positioning is a bit off for me:

act-sticky

wioux:
I still see that issue with the tool title getting mixed up over multiple lines.

lschatzkin:
> The sticky positioning is a bit off for me:
> 
> act-sticky

Can I come see it on your screen? I'm not getting the same behavior.

lschatzkin:
> I still see that issue with the tool title getting mixed up over multiple lines.

Hmm, I'm not seeing this... is there a particular type of action that you see this on? Or all of them?

lschatzkin:
@wioux I can't duplicate your problem with the line height in my sandbox, but hopefully this will solve the problem. Can you check?

wioux:
The overlap issue is fixed, though we should circle back and remove the new line-height rule as part of #550. Using a relative value `line-height: 1` would cover both these situations:

```css
h3 {
  font-size: 29px;
  line-height: 29px;
}

h3.tool-title {
  font-size: 4rem;
  line-height: 4rem;
}
```

lschatzkin:
> The overlap issue is fixed, though we should circle back and remove the new line-height rule as part of #550. Using a relative value `line-height: 1` would cover both these situations:
> 
> ```css
> h3 {
>   font-size: 29px;
>   line-height: 29px;
> }
> 
> h3.tool-title {
>   font-size: 4rem;
>   line-height: 4rem;
> }
> ```

Agreed.

lschatzkin:
Are folks okay with my merging this?

app/assets/stylesheets/action_page.scss
```diff
@@ -14,64 +31,48 @@
...
+      color: white;
       text-transform: uppercase;
       text-align: center;
 
       .tool-title {
-        text-transform: none;
-        font-size: 1.2em;
+        color: white;
+        font-family: $league;
+        font-size: 4rem;
```
wioux: The mockup seems to anticipate a short tool title like this:

Screenshot_2019-06-26_15-07-21

but we allow those titles to be customized, so we have [some actions](https://act.eff.org/action/california-demand-community-control-of-police-surveillance-2b7ce8a3-0f59-4164-8d6a-1adb8fc306f4) where instead it'll look like

Screenshot_2019-06-26_14-52-49

Something closer to 2rem / not uppercased might be better?

esoterik:
i'm still seeing the nav bar clipping the tools and the large space between the bottom of the action and the footer

app/assets/stylesheets/action_page.scss
lschatzkin: Asking design folks.

lschatzkin:
> i'm still seeing the nav bar clipping the tools and the large space between the bottom of the action and the footer

Would you be okay if I did that in another ticket?

esoterik:
> > i'm still seeing the nav bar clipping the tools and the large space between the bottom of the action and the footer
> 
> Would you be okay if I did that in another ticket?

sgtm!

lschatzkin:
> > > i'm still seeing the nav bar clipping the tools and the large space between the bottom of the action and the footer
> > 
> > 
> > Would you be okay if I did that in another ticket?
> 
> sgtm!

Okay, it exists as #555

app/assets/stylesheets/action_page.scss
lschatzkin: It appears the Activism might be okay with this always saying "Take Action". So I'm going to punt that issue for now.

lschatzkin:
I'm going to work on the sticky tools box in #555, going to work on relative line heights with #550 , and I'm going to merge this branch.

PR #553, opened 2019-06-25
Move congress forms integration to backend

vbrown608:
Laying the groundwork for moving our congressforms integration to the backend. This will eventually replace `congress_forms.js` and `congress_message.js`. I tried to follow the API of the congress forms gem somewhat in case we'd like to include that in the Action Center at some point in the future.

This will probably move around a bit as I figure out how to things need to be rendered in the view and experiment a bit more with what kind of submissions the `/fill-out-form/` endpoint expects.


vbrown608:
Thanks for the review @sky! I'm going to roll these into the rest of the feedback I get on #565. I took your advice about including `ActiveModel::Model` and I think it was good!

PR #556, opened 2019-06-26
491 action page banner

lschatzkin:
I chose the background image for the banner image, as I thought the dimension would be better than the featured image. A few other style fixes as well.


app/views/action_page/show.html.erb
```diff
@@ -1,12 +1,16 @@
...
+
 <% end -%>
 <div class="top-section">
   <div class="container">
     <div class="row">
       <div>
+        <h5 class="category">Category Placeholder</h5>
         <h1><%= @actionPage.title -%></h1>
+        <div class="image">
+          <img src="<%= @actionPage.background_image.url -%>">
```
vbrown608: We should use a helper to render the image here so that the url isn't hard-coded.
```suggestion
          <%= image_tag @actionPage.background_image -%>
```

lschatzkin: Cool, I'll change that.

PR #561, opened 2019-06-27
Render contact congress form

vbrown608:
Building on #553 - this PR renders the form 

Still todo there - some logic around pre-populating fields and looking up the correct bioguide IDs for the campaign.

This can be tested by visiting, eg, /congress_messages/new?street_address=815+Eddy+Street&zipcode=94109&congress_message_campaign_id=139


spec/lib/congress_forms_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+        { "value" => "$NAME_LAST" },
+        { "value" => "$ADDRESS_STATE", "options_hash" => {
+            "CALIFORNIA" => "CA",
+            "NEW YORK" => "NY"
+        } }
+      ])
+    }
+
+    let(:input) { {
+      "$NAME_FIRST" => "Willow",
```
lschatzkin: I'm not sure what this is for?

lschatzkin:
When I try to go to the URL listed above, I get: Couldn't find CongressMessageCampaign with 'id'=139. I looked at it with id=138, which had nothing filled in, which is what I thought would happen.

I'm seeing the message area as a text field, but I think it should be text area... but I think that may be in another branch?

spec/lib/congress_forms_spec.rb
vbrown608: Yep! Everything under `/spec` and `/features` is for testing.

In the past we've put browser tests in `/features` and used [Cucumber]https://cucumber.io/, but going forward we're putting all our testing in `/spec` (including browser tests) and dropping that tool.

PR #563, opened 2019-06-28
558 new nav

### LINKED ISSUE #558: Nav changes per Hugh
Mockup coming.

lschatzkin:
Redoing nav, based on Hugh's input.

Added a modal-type behavior so user can click anywhere outside menu to close menu.

closes #558


esoterik:
It'd be nice to keep the same mobile behavior as before, I think (the slide-out full length menu rather than the short two links popping up). 

I also wonder if the menu should be hidden (on desktop) when not hovering over it  or the hamburger. To me it feels unexpected when it stays open and a little clunky / intuitive to have to click the hamburger again to close it, but I know there are accessibility concerns with hover-only content. Maybe clicking outside the div with the links should also close the menu?

lschatzkin:
> It'd be nice to keep the same mobile behavior as before, I think (the slide-out full length menu rather than the short two links popping up).
> 
> I also wonder if the menu should be hidden (on desktop) when not hovering over it or the hamburger. To me it feels unexpected when it stays open and a little clunky / intuitive to have to click the hamburger again to close it, but I know there are accessibility concerns with hover-only content. Maybe clicking outside the div with the links should also close the menu?

I discussed the mobile with Hugh, and he likes the menu being the same on mobile and desktop.

I'm not sure about clicking outside the drop down to close. It seems I've mostly seen that with modals. Do you think users would do that? Maybe let's discuss with design folks.

app/assets/javascripts/application/nav-close.js
```diff
@@ -0,0 +1,11 @@
+$(document).click(function(event) {
+  $navmenu = $('#nav-menu');
+  $target = $(event.target);
+  if(!$target.closest('#nav-menu').length &&
+  $('#nav-menu').is(":visible")) {
+    $('#nav-menu').css('display', 'none');
+  }
+  else {
+      $('#nav-menu').css('display', 'inline-block');
+  }
```
vbrown608: This file would be easier to read with more conventional whitespace, eg:
```
if (!$target.closest('#nav-menu').length && $('#nav-menu').is(":visible")) {
  $('#nav-menu').css('display', 'none');
} else {
  $('#nav-menu').css('display', 'inline-block');
}
 ```
For javascript, we should use single quotes consistently for strings. In the medium term, we should add a javascript linter to the action center so we can catch these kinds of things automatically.

File: app/assets/javascripts/application/nav-close.js [hunk 2]
```diff
@@ -0,0 +1,11 @@
+$(document).click(function(event) {
+  $navmenu = $('#nav-menu');
+  $target = $(event.target);
```
vbrown608: These variables are getting global scope - they should be prefixed with `var`.

wioux:
The menu pops up in an unexpected spot:

Screenshot_2019-07-18_15-18-41

wioux:
The menu is broken if you click anywhere else on the page before opening it.

app/assets/javascripts/application/nav-close.js
lschatzkin: Oh, this is some old code. I'm not actually using this, I'm only using css. I'll delete it.

app/assets/javascripts/application/nav-close.js
lschatzkin: See above.

lschatzkin:
> The menu is broken if you click anywhere else on the page before opening it.

:(

lschatzkin:
Okay, thanks @wioux for finding that problem. It turns out the legacy jquery code was obstructing the css from working. @vbrown608 ready for re-review.

lschatzkin:
> The menu pops up in an unexpected spot:
> 
> Screenshot_2019-07-18_15-18-41

Oops! Forgot to fix this, doing it now.

lschatzkin:
Ready for review again.

vbrown608:
:dancer: works great!

Issue #505, opened 2019-05-22
Add Thank you text to last page of form


lschatzkin:
See mockup: there should be social icons too.

PR #564, opened 2019-07-09
Add new thank you flow for actions

### LINKED ISSUE #502: Add illustration to thank you page

esoterik:
Closes #502, #504, and #505

The thank you section is pretty much unstyled at the moment. It also is not tested for congress message actions -- I kept getting a CORS error after entering an address and hitting "Find your reps".

There are some new test failures in this PR but I'll handle them in the big fix-all-the-tests branch that I've been working on. As far as I can tell, the new failures are entirely a result of visual changes, not any actual functionality breaking.


lschatzkin:
This is great.
For the petitions, I'm seeing both the Signed section and the Thank you section. Is that correct, and I need to fix the Signed section, or should only the Thank you appear?
signed
Also, I never seem to get to a Thank You page for the Call Tool, not sure if that's correct.

esoterik:
okay, changed the behavior for call and petition actions!

lschatzkin:
Now that I'm looking at it again, I think we should have the thank you appear at the bottom of the tweet action as well, since that is the last screen the user sees. Sorry I didn't notice this sooner.
tweet action

esoterik:
hmm, for me twitter opens in a new, small window and the action center stays visible. It seems potentially confusing to show the "thank you" section before the user does anything

lschatzkin:
Okay, let's leave it as is. Merging!

PR #565, opened 2019-07-10
Backend congress forms integration

vbrown608:
This PR moves our Congress Forms API integration from Javascript (mostly in `congress_forms.js` and `congress_message.js`) to the backend.

I'm going to mark a few places where I'd like to do a little follow-up work after this lands. This PR was getting huge so I wanted to hurry up and get it open.

I'll also open a follow-up PR with deletions - I marked a few places in the code as reminders of some odds and ends that can come out. 

My intention is that `lib/congress_forms.rb` wraps our interactions with the Congress Forms API while `app/models/congress_message.rb` describes a congress message submission by an Action Center user.


app/controllers/action_page_controller.rb
```diff
@@ -38,9 +38,6 @@ def embed
 
   def embed_iframe
     @css = params[:css] if params.include? :css
-    @target_bioguide_ids = params[:bioguide_ids] if params.include? :bioguide_ids
```
vbrown608: We added this functionality for checkyourreps, which we aren't using anymore. I'm dropping it.

app/controllers/congress_messages_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+    @message = CongressMessage.new_from_lookup(location, params[:message], @campaign, forms)
+    render partial: "form"
+  end
+
+  def create
+    @message = CongressMessage.new(congress_message_params.merge(campaign: @campaign))
+    @message.forms = CongressForms::Form.find(params["bioguide_ids"].split)
+
+    if @message.submit
+      @name = user_params[:first_name] # for deliver_thanks_message
```
vbrown608: I'd like to start moving some of the helpers in `tools_controller.rb` to the tooling concern. Many of them rely really heavily on instance variable being set, which I don't think is ideal, but I didn't want to move too much more around for this PR.

app/controllers/congress_messages_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+    # In Rails 5.1 we can do params.permit(common_attributes: {}, member_attributes: {})
+    params.permit.tap do |p|
+      p[:common_attributes] = params[:common_attributes] && params[:common_attributes].permit!
+      p[:member_attributes] = params[:member_attributes] && params[:member_attributes].permit!
+    end
+  end
+
+  def user_params
+    params.require(:common_attributes).permit.tap do |p|
+      common = params[:common_attributes]
```
vbrown608: This mapping to and from Congress Forms-style field name formatting also happens in `CongressMessage::new_from_lookup`. It might be nice to do it all internally in that class and use friendlier field/param names out here. 

In general, I wasn't totally sure about the best way to represent filled in values there. I went with `common_attributes` and `member_attributes` to mirror the way `attributes` returns a hash, but I'm open to other ideas.

app/controllers/congress_messages_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+    @message = CongressMessage.new(congress_message_params.merge(campaign: @campaign))
+    @message.forms = CongressForms::Form.find(params["bioguide_ids"].split)
+
+    if @message.submit
+      @name = user_params[:first_name] # for deliver_thanks_message
+      track_action
+      deliver_thanks_message
+      render partial: "tools/share"
+    else
+      render plain: I18n.t(:submission_failed, scope: :congress_forms), status: :bad_request
```
vbrown608: In the current version, `500`s from the Congress Forms API submit endpoint fail silently. This shows an error, but we might want to preserve the old behavior since the attempted fills will still hopefully be saved in the delayed job queue there. Thoughts?

spec/requests/congress_messages_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+      and_return(status: 200, body: forms_body.to_json)
+  end
+
+  before do
+    allow(SmartyStreets).to receive(:get_location).and_return(location)
+    stub_congress_forms_find_with_two_reps
+  end
+
+  describe "#new" do
+    subject do
```
esoterik: I think it would be more clear if the `subject` blocks were changed to methods with descriptive names

spec/requests/congress_messages_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+          "$NAME_PREFIX": "Mrs.",
+          "$ADDRESS_STATE_POSTAL_ABBREV": "CA",
+          "$MESSAGE": "Impeach Mayor Richard Wilkins III",
+          "$TOPIC": "JU"
+        },
+        campaign_tag: "a campaign tag"
+      })
+    end
+
+    it "returns an error when validation fails" do
```
esoterik: is there a way we can check for this error message rather than just checking that it didn't `post`?

spec/features/congress_message.rb
```diff
@@ -0,0 +1,79 @@
...
+    select "Mr.", from: "member_attributes_C000880__NAME_PREFIX"
+    select "Education, Science & Technology", from: "member_attributes_C000880__TOPIC"
+    select "Education", from: "member_attributes_A000360__TOPIC"
+    click_button "Submit"
+
+    expect(page).to have_content "Now help spread the word"
+    expect(partner.subscriptions.count).to eq 1
+    expect(WebMock).to have_requested(:post, CiviCRM::supporters_api_url).
+      with(body: hash_including({
+        data: '{"contact_params":{"email":"mrgiles@sunnydale.edu","first_name":"Rupert","last_name":"Giles","source":"action center congress message :: Sample Action Page","subscribe":true,"opt_in":true},"address_params":{"city":"Sunnydale","state":null,"street":"The Library","zip":"94109","country":null},"phone":null}'
```
esoterik: do we want to be this specific here? I'd assume that this is pretty stable, but this expectation seems frail. Could we use a hash to fill in the form values and this string to mitigate some of that?

app/assets/javascripts/application/tools/congress_message.js
```diff
@@ -1,4 +1,50 @@
 $(document).on("ready", function() {
+  $(".congress-message-rep-lookup, #congress-message-create").on("ajax:beforeSend", function() {
+    submit_in_progress($(this));
+  });
+
+  $(".congress-message-rep-lookup").on("ajax:complete", function(xhr, data, status) {
+    $form = $(this);
```
wioux: `$form`/`$phone` are given global scope but it looks like they're only needed within this file.

app/controllers/congress_messages_controller.rb
```diff
@@ -0,0 +1,101 @@
+class CongressMessagesController < ApplicationController
+  include Tooling
+
+  before_action :set_congress_message_campaign
+  before_action :update_user, only: :create
+  before_action :subscribe_user, only: :create
+
+  rescue_from SmartyStreets::AddressNotFound, with: :address_not_found
```
wioux: :+1: I like the use of `rescue_from` here.

app/controllers/congress_messages_controller.rb
wioux: Yes, failed messages are retried periodically, in case an update is made in [unitedstates/contact-congress](https://github.com/unitedstates/contact-congress) which will let them succeed. Submissions can also be rejected for reasons that aren't fixable, but we can't differentiate the two failure types currently.

Since these are usually form messages written by EFF I think it's not the worst thing to swallow errors and make a best effort at delivering the statement later. It's more problematic if the message is fully authored by the user, as it was on democracy.io, or sometimes might be on act.eff.org. In that case it might be worth showing some kind of notification and providing a link directly to the senate.gov contact page?

app/assets/javascripts/application/tools/congress_message.js
```diff
@@ -1,4 +1,50 @@
...
+  function show_error(error, form) {
+    form.find(".progress-striped").hide();
+    form.find("input[type=submit]").show();
+    form.find(".alert-danger").remove();
+    $("#errors").append($('<div class="small alert alert-danger help-block">').text(error));
+    $("input,textarea,button,select", form).removeAttr("disabled");
+  }
 
   var emailsSent = 0;
   var allEmailSent = function () {
```
wioux: Is it safe to remove the old code for this tool? [L230](https://github.com/EFForg/action-center-platform/pull/565/files#diff-29449e46fbc2722e2392c357d670ffafL230) is throwing an error when I look up reps (though it doesn't affect functionality).

wioux:
Thanks for taking this on, looks like a big lift. The only major thing I see is that we need some way to deal with long-running synchronous requests to congress_forms_api on the backend. The POSTs to /fill-out-form can take a while and they'll tie up a webserver thread while they wait.

If it's tricky to issue the requests asynchronously from action center, we could pretty easily have congressforms shove the messages directly into the delayed_job worker on its end, and then respond immediately. In either case though we'll probably lose the ability to discriminate between successful/failed messages.

We should also restore the UI change that happens while the message is submitting (display rep names with a progress/wait animation). It's not clear when you hit the submit button that anything has happened, though this problem is exacerbated by the long-running request issue.

vbrown608:
@wioux yeah, I was thinking that would be a nice side effect of deciding we don't mind swallowing the error :) I'll look at some options for submitting asynchronously.

app/controllers/congress_messages_controller.rb
vbrown608: Maybe we could send an e-mail notifying the user and linking them to the senate.gov page. That way submission could still happen asynchronously.

app/controllers/action_page_controller.rb
```diff
@@ -123,6 +120,7 @@ def set_action_display_variables
     if @email_campaign and !@email_campaign.topic_category.nil?
       @topic_category = @email_campaign.topic_category.as_2d_array
     end
+    # @TODO remove
```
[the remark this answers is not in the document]
vbrown608: yeah, I left a few reminders to myself of code that needs to come out when I open a big delete stuff PR

lib/congress_forms.rb
```diff
@@ -1,21 +1,139 @@
...
+    def eql?(other)
+      self == other
+    end
+
+    def hash
+      [max_length, value, options_hash].hash
+    end
+
+    def options_hash
+      if @value == "$ADDRESS_STATE_POSTAL_ABBREV"
```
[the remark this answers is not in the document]
vbrown608: Since this just returns a value based on a string comparison, I'm not sure the performance improvement would be worthwhile. I think we could add that behavior if setting up the options_hash becomes more complex.

spec/factories/congress_message.rb
```diff
@@ -0,0 +1,22 @@
+FactoryGirl.define do
+  form1 = CongressForms::Form.new("C000880", [
```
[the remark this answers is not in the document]
vbrown608: My understanding is that bindings that are created inside a block disappear after that block ends. `form1` is undefined if I try to access it in a spec.

spec/features/congress_message.rb
vbrown608: Oof yeah I wrestled with this a bit. I agree that it's very fragile. The API we're submitting to expects the data field to be a JSON-encoded string (so our submission is sort of being form encoded twice) which makes it hard to use the flexible matchers that `WebMock` provides. We can use `hash_including` the specify the data field but not sub fields because the data field is a JSON-encoded string rather than a hash.

I do want to make sure all the relevant user information is getting passed to CiviCRM, since there's some complexity around how we extract it from various tools that structure their forms differently. 

Building the string out from a hash feels like it would duplicate the logic in the CiviCRM service object. Maybe I can put this mock on that instead of mocking the request.

spec/models/topic_category_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      expect(subject.best_match options).to eq "Vampire2"
+    end
+
+    it "ignores case, whitespace, and punctuation when matching" do
+      options = ["Vampire3", "VamPire 2!"]
+      expect(subject.best_match options).to eq "VamPire 2!"
+    end
+
+    after do
+      FactoryGirl.reload
```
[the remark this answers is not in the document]
vbrown608: This is resetting the sequence numbers  after each test - otherwise I get Vampires 5-8 in whatever test runs second.

Thoughtbot recommends naming models explicitly rather than relying on sequence numbers when they're going to be referenced in test assertions, but it considering the tree of the topic-related models that needs to get created here it seemed worthwhile to rely on the factories as they're written.

I could add a comment explaining what's going on here.

vbrown608:
The Action Center is already configured to support delayed jobs, so I went ahead and added them to the Action Center's queue.

app/views/congress_messages/_form.html.erb
```diff
@@ -0,0 +1,43 @@
...
+
+    <%= render "tools/newsletter_signup",
+               email: false, location: false, privacy_notice: false %>
+    <%= render "tools/save_my_info_option", info: %w(name address) %>
+  <% end %>
+
+  <% @message.forms_minus_common_fields.each do |form| %>
+    <fieldset class="congressForms-legislator-fields">
+      <h3 class="legislator-label">
+        <em><%= CongressMember.find_by(bioguide_id: form.bioguide_id).full_name %></em>
```
lschatzkin: There's no <em> in the new style.

app/views/congress_messages/_form.html.erb
```diff
@@ -0,0 +1,43 @@
...
+      </h3>
+      <% form.fields.each do |field| %>
+        <div class="form-group">
+          <%= congress_forms_field(field, @campaign,
```
lschatzkin: I believe the new version is supposed to have a checkbox next to each legislator's name so they can choose who they submit to. I can try to add this in when this branch is ready if you like.

vbrown608:
@lschatzkin thanks for flagging these issues! This branch is going to be merged into master, so it should preserve the current action center styles. I'll create a ticket for a new PR to upgrade this markup to reflect the redesign.

app/controllers/congress_messages_controller.rb
wioux: :+1: I like that solution.

app/assets/javascripts/application/tools/congress_message.js
```diff
@@ -4,12 +4,12 @@ $(document).on("ready", function() {
   });
 
   $(".congress-message-rep-lookup").on("ajax:complete", function(xhr, data, status) {
-    $form = $(this);
+    var $form = $(this);
```
wioux: `$form` is referenced in the other callback function (below), so it looks like this needs to be declared further up -- though I think it would also be fine if the `show_error` function found the form itself and didn't need it passed in as an argument.

vbrown608:
Approved by @wioux

PR #573, opened 2019-07-15
572 kill sticky tools

### LINKED ISSUE #572: change Take Action form to not floating down the page

lschatzkin:
closes #572


wioux:
Looks like this should be a PR against the redesign-2019 branch.

lschatzkin:
> Looks like this should be a PR against the redesign-2019 branch.

Indeed.

PR #574, opened 2019-07-15
Fix some tests for the user-facing redesign

esoterik:
Doesn't fix all of them, but it's an okay start.

I'm fairly certain that the remaining poltergeist overlapping errors are caused by the tools div existing within a row, which should probably be addressed in the styles, not the tests.


lschatzkin:
> I'm fairly certain that the remaining poltergeist overlapping errors are caused by the tools div existing within a row, which should probably be addressed in the styles, not the tests.

Tools is being revamped in another branch, so maybe that will fix the remaining errors.

lschatzkin:
LGTM!

PR #575, opened 2019-07-16
Action Cloning

### LINKED ISSUE #540: [Action Cloning] Add basic action page cloning
Part 1 of the action page cloning feature.

This issue encompasses getting the basic model-level cloning ready: just getting `ActionPage` attributes from one existing record into a new, not persisted instance (no associations). No front-end or controller work just yet!

esoterik:
Closes #540

Extremely basic action cloning. Doesn't handle associations super well yet (category is fine; any action-related associations like a tweet are less fine).

### DIFF
--- a/lib/action_cloner.rb
+++ b/lib/action_cloner.rb
@@ -3,24 +3,17 @@ module ActionCloner
     clone = original.dup
     clone.published = false
     clone.archived = false
-    clone_campaign(original, clone)
+    clone_tools(original, clone)
   end
 
   private
 
-  def self.clone_campaign(page, clone)
-    campaign_sym, model = determine_campaign(page)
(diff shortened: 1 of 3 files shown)


vbrown608:
Thanks for your writing such thorough tests! I think they were really effective - I felt confident making implementation changes and expecting that the functionality was preserved. Also appreciating the thought you put into which attributes need to preserved, which need to be updated, which need to be deep-copied, etc.

It turns out ActiveRecord provides a method called `dup`, which handles cloning a record (but not its associations). 

For cloning the campaigns/tools, the action center currently supports multiple tools on an action. I'm not sure it ever appears outside of testing, but I updated the logic to iterate over all possible tools instead of selecting a single one just in case.

PR #578, opened 2019-07-16
Fix libssl in docker build

esoterik:
Cherrypicked @wioux 's fix on the admin redesign branch to master so our builds work again.


esoterik:
update: the build failures were caused by the ruby-2.5 image updating to debian buster. On buster, poltergeist couldn't find the libssl libraries, and it was proving to be a more involved issue. This downgrades us to a debian stretch based image for now so our builds work again.

PR #583, opened 2019-07-24
580 thank you

lschatzkin:
Styling, social share links and all things thank you.


app/helpers/action_page_helper.rb
```diff
@@ -38,6 +38,9 @@ def email_friends_url(action_page)
     "mailto:?subject=#{u subject}&body=#{u message}"
   end
 
+  def link_share_url(action_page)
```
vbrown608: What's the purpose of this helper? Seems like maybe we're getting the share url in javascript in `copy-url.js`.

app/views/tools/_petition.html.erb
```diff
@@ -104,7 +104,9 @@
     </div>
   </div>
 </div>
-
+</div>
+</div>
+</div>
```
vbrown608: What's the reason for these additional closing div tags? It looks like they might not have matching opening tags in the template.

app/views/tools/_thank_you.html.erb
```diff
@@ -2,27 +2,36 @@
...
+      <hr>
+      <p> EFF is supported by donations from folks like you. We appreciate anything you can give! </p>
+      <a class="eff-red-button" href="https://supporters.eff.org/donate">Donate</a>
+      <% if @actionPage.enable_tweet && @actionPage.tweet %>
       <% if @actionPage.tweet.targets.present? && (@actionPage.tweet.targets.size > 1) %>
-        <span id="tweet-back">Want to tweet at others? Go back</span>
+      <span id="tweet-back">Want to tweet at others? Go back</span>
+      <% end %>
       <% end %>
-    <% end %>
```
vbrown608: Indenting within `if` statements would be helpful for readability. Something like:
```
      <% if @actionPage.enable_tweet && @actionPage.tweet %>
        <% if @actionPage.tweet.targets.present? && (@actionPage.tweet.targets.size > 1) %>
          <span id="tweet-back">Want to tweet at others? Go back</span>
          <span id="tweet-back">Want to tweet at others? Go back</span>
        <% end %>
      <% end %>
```

app/helpers/action_page_helper.rb
lschatzkin: You're right, that's a remnant of another thought path I went down.

app/views/tools/_petition.html.erb
lschatzkin: That was a hack I forgot to remove, removing now.

app/views/tools/_thank_you.html.erb
lschatzkin: You speak the truth, my friend.

Issue #580, opened 2019-07-18
Thank you styling


lschatzkin:

lschatzkin:
This is included in design-cleanup3

PR #584, opened 2019-07-26
Design cleanup3

lschatzkin:
Another round of design tweaks.

Merge after #573 kill sticky tools and #580 thank you styling
This has #573 and #580 already merged into it, because I couldn't move forward without those.


app/views/tools/_petition.html.erb
```diff
@@ -101,10 +102,13 @@
             <%= render 'tools/loading' -%>
           <% end %>
       </div>
+      </form>
     </div>
   </div>
 </div>
-
+</div>
```
vbrown608: Looks like some extra closing divs may have made it in here.

PR #585, opened 2019-07-29
Consolidate place data

vbrown608:
Now that we we don't need it for front-end congress forms, we can remove `states.js`. I aggregated a couple of the other places where we've got arrays/hashes of places into `places.rb`.


app/models/signature.rb
```diff
@@ -61,7 +61,7 @@ def to_csv_line
   end
 
   def state_symbol
-    us_states_hash[state]
```
wioux: Nice, looks like this model can stop including `ApplicationHelper` now. :confetti_ball:

vbrown608: Oo great!

PR #586, opened 2019-07-29
Update admin nav

esoterik:
~~Not 100% sure on what to do with the non-link in the nav (Advanced Config section header) -- the wireframes have it as the same color, but would that be confusing? Left it alone, can update to make it the same if necessary.~~
this was based on old wireframes, will push changes soon

### DIFF
--- a/app/assets/stylesheets/admin/layout.scss
+++ b/app/assets/stylesheets/admin/layout.scss
@@ -29,6 +29,50 @@ header {
     margin-bottom: 0;
     margin-left: 1.5rem;
   }
+
+  .top-nav {
+    display: inline-block;
+    width: 100%;
+    @include medium {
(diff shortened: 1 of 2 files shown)


esoterik:
okay, updated!

wioux:
Thanks for the mobile styles! Looks good, just a couple small notes:

  * I've been sticking to em/rem units rather than px for font sizes (16px = 1rem).

  * There is a typo in the font-family name, it should be Mon*t*serrat -- you could also use the variable `$sans-serif` to fix this. 

  * The `<li>` cells could use some padding, when the window width is less than about 1100px they bunch up Screenshot_2019-07-30_11-06-40

esoterik:
made those changes + made the nav text align with the logo and page title!

PR #590, opened 2019-07-31
Admin Redesign: action page table menu

esoterik:
A couple of unrelated commits from master made it in due to how I grabbed the cloning functionality, but they're not causing merge conflicts so I think it's okay!

This adds a menu to the admin actions table that allows admins to view, edit, duplicate, or destroy actions.


esoterik:
resolved conflicts post merging #591 !

PR #591, opened 2019-07-31
Add action previews to new admin interface

esoterik:


wioux:
This wasn't working for me initially, I think because of an interaction between the `target=_blank` attribute on the link and the `window.open()` call. The `window.open()` call was returning null, possibly getting blocked because the link itself had just opened a new window? Two ways I found to fix it are

  * Calling `.preventDefault()` on the click event
  * Using the [`windowName` parameter](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#Parameters) and passing the same value as the link's `target` attribute (for this to work though, target cannot be `_blank`)

esoterik:
weird! Removed `target="_blank"` -- it was a holdover from the previous implementation anyway.

config/routes.rb
```diff
@@ -102,7 +102,7 @@
       get :unpublish
       get :destroy
       post 'update_featured_pages', :on => :collection
-      patch :preview
+      get :preview
```
wioux: The method here I think had to do with browsers and webservers having various limits on how long a url (including the query string) can be. [Fastly](https://docs.fastly.com/guides/debugging/resource-limits#request-and-header-limits) for example will send back an error if the url exceeds 8KB. The preview I just checked requested a 5KB long url, which is getting up there, so it'd be safer to leave this as a post/patch so the form data is sent in the request body instead.

PR #592, opened 2019-08-01
Adding text to admin

lschatzkin:
Text updates to the admin areas that are currently built out. Added a css tooltip

To do:

- congress actions
- add logo
- mailing list
- status options

### DIFF
--- a/app/assets/stylesheets/admin/admin.scss
+++ b/app/assets/stylesheets/admin/admin.scss
@@ -47,3 +47,38 @@
 #epiceditor {
   border: 1px solid #ccc;
 }
+
+
+/* Tooltip container */
+.tooltip {
+  color: $gray;
+  border: 2px solid $gray;
+  border-radius: 1rem;
+  display: inline-block;
+  font-size: 1.1rem;
+  font-weight: 800;
+  padding: 0 7px;
+  position: absolute;
+  right: 0;
+  top: 0;
+  .tooltiptext {
+    background-color: black;
+    border-radius: 0.5rem;
+    color: white;
+    font-size: 1rem;
+    font-weight: 400;
+    left: 2rem;
+    padding: 1rem;
+    position: absolute;
+    text-align: left;
+    visibility: hidden;
+    width: 20rem;
+    z-index: 1;
+  }
+  &:hover {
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/admin/base.scss
```diff
@@ -1,6 +1,42 @@
...
\ No newline at end of file
+}
+
+/* Tooltip container */
+.tooltip {
+  color: $gray;
+  border: 2px solid $gray;
+  border-radius: 1rem;
+  display: inline-block;
+  font-size: 18px;
```
wioux: I've been sticking to em/rem units for font-size (16px = 1rem).

wioux:
There's not a super clear method to the style organization now, but I'm trying to keep most of the general rules in admin.scss. Could you move the tooltip styles there? I've been putting variables and colors in mixins.scss, so that's where $eff-red/$gray should go.

Thanks for starting on this!

app/assets/stylesheets/admin/base.scss
lschatzkin: Oh, me too, thanks for catching that.

lschatzkin:
Is there a way to share the variables.scss file in /application, or do we need to just recreate them? I'm thinking about the colors but it's not a big deal.

I keep usually keep global variables in variables.scss, I'm happy to reorganize the files if that works for you.

app/assets/stylesheets/admin/admin.scss
```diff
@@ -47,3 +47,38 @@
 #epiceditor {
```
wioux: Aahh I didn't realize this file still exists, I plan to delete it. The admin styles are in `app/assets/stylesheets/admin.scss`.

lschatzkin: oops!

wioux:
I'd like to keep the themes separate and the number of variables on the admin side pretty small. The light theme does well with a smaller color palette IMO.

lschatzkin:
> I'd like to keep the themes separate and the number of variables on the admin side pretty small. The light theme does well with a smaller color palette IMO.

Okay, works for me.

PR #594, opened 2019-08-01
Update redesign branch for backend congress forms

vbrown608:
Update the redesign branch to work with backend congress forms. This really just involved removing the old div that held all the data required to generate the forms on the frontend.

The markup for actual reps' forms (previously generated in js) can now be found in `/app/views/congress_messages/_form.html.erb`. It might need some updates to work well with the new styles.


lschatzkin:
When I test the congress forms I get a socket error, is that expected at this point?

PR #596, opened 2019-08-02
560 category

### LINKED ISSUE #560: Add Category to Action page template, home page
Right now there's just a placeholder in the action page, nothing on home.

lschatzkin:
adds category to home page, index page, action pages

closes #560


app/views/action_page/index.html.erb
```diff
@@ -10,6 +10,10 @@
       <div class="row">
         <% array_of_3_actions.each do |actionPage|%>
             <div>
+              <% if Category.exists? %>
+              <h5 class="category">
+                <%= Category.where(id: actionPage.category_id).pluck(:title).first; %></h5>
+              <% end %>
```
vbrown608: Templates should be indented after `if` statements, like 
```
              <% if Category.exists? %>
                <h5 class="category">
                  <%= Category.where(id: actionPage.category_id).pluck(:title).first; %>
                </h5>
              <% end %>
```

app/views/action_page/index.html.erb
```diff
@@ -10,6 +10,10 @@
       <div class="row">
         <% array_of_3_actions.each do |actionPage|%>
             <div>
+              <% if Category.exists? %>
```
vbrown608: `Category` refers to the overall class of categories. 

We want to get the category for current action page (stored in the `actionPage` variable inside this loop). ActiveRecord makes that easy - we can just call `actionPage.category`. 

To check if it's not nil or empty, we do `actionPage.category.present?` I've found this table really helpful for these types of checks: https://stackoverflow.com/a/26877095

Putting it together, you'd have something like
```
<% if actionPage.category.present? %>
  <%= actionPage.category.title %>
<% end %>
```

lschatzkin: Weird, I swear I tried that and it didn't work. Oh well!

app/views/action_page/index.html.erb
```diff
@@ -10,6 +10,11 @@
       <div class="row">
         <% array_of_3_actions.each do |actionPage|%>
             <div>
+              <h5>
```
vbrown608: This new version will result in an empty `h5` if `category` isn't present. It would be better to put the tag inside the conditional, like
```
<% if actionPage.category.present? %>
  <h5><%= actionPage.category.title %></h5>
<% end %>
```

PR #597, opened 2019-08-05
548 fix tooltips

### LINKED ISSUE #548: Fix tooltips
not using bootstrap

lschatzkin:
replacing bootstrap popovers with css on the front end.

TODO: Still need to do the same for admin.

closes #548


vbrown608:
Magic!

PR #598, opened 2019-08-06
Add more filters to the admin dashboard

esoterik:
Also adds an author association for action pages.

### DIFF
--- a/app/controllers/admin/action_pages_controller.rb
+++ b/app/controllers/admin/action_pages_controller.rb
@@ -18,11 +18,8 @@ class Admin::ActionPagesController < Admin::ApplicationController
   allow_collaborators_to :index, :edit
 
   def index
-    @categories = Category.all.order(:title).map { |c| [c.title, c.id] }
-    @authors = User.joins(:action_pages).order(:last_name).map do |a|
-      label = a.name.blank? ? a.email : a.name
-      [label, a.id]
-    end
+    @categories = Category.all.order(:title)
+    @authors = User.authors.order(:last_name)
     @actionPages = filter_action_pages
 
     if request.xhr?
--- a/app/controllers/admin/action_pages_controller.rb
+++ b/app/controllers/admin/action_pages_controller.rb
@@ -18,6 +18,11 @@ class Admin::ActionPagesController < Admin::ApplicationController
   allow_collaborators_to :index, :edit
 
   def index
+    @categories = Category.all.order(:title).map { |c| [c.title, c.id] }
+    @authors = User.joins(:action_pages).order(:last_name).map do |a|
+      label = a.name.blank? ? a.email : a.name
+      [label, a.id]
+    end
     @actionPages = filter_action_pages
 
     if request.xhr?
@@ -36,7 +41,7 @@ class Admin::ActionPagesController < Admin::ApplicationController
   end
 
   def create
-    @actionPage = ActionPage.new(action_page_params)
+    @actionPage = ActionPage.new(action_page_params.merge(author: current_user))
 
     if @actionPage.save
       redirect_to action_page_path(@actionPage)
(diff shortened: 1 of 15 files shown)


app/queries/action_page_filters.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  private
+
+  NAMED_SCOPES = %i(type status).freeze
+  VALID_FILTERS = %i(type status author category).freeze
+
+  attr_accessor :relation, :filters
+
+  def default_date_range?(start_date, end_date)
+    start_date == end_date && start_date == Time.zone.today
```
esoterik: I'm not happy with how this is handled -- basically this checks to see if the date range given is the default, and doesn't run the date query if it is. 

Ideally the date picker would be blank by default and it would be easier for the user to clear it out entirely, but I wasn't sure how to do that.

spec/queries/action_page_filters_spec.rb
```diff
@@ -0,0 +1,48 @@
+require "rails_helper"
+
+describe ActionPageFilters do
```
esoterik: these tests aren't pretty, but testing this new object requires detailed data. 

I'm not sure if there are any good ways to clean them up, but it would be nice to do that.

spec/requests/admin/action_pages_spec.rb
```diff
@@ -64,12 +64,6 @@
       expect(response.body).to include("borderpetition")
       expect(response.body).to include("bordertweet")
       expect(response.body).not_to include("privacypetition")
-
-      xhr :get, "/admin/action_pages?q=border+surveil&action_type=petition"
```
esoterik: Removed this request spec because the filters are now tested elsewhere.

app/queries/action_page_filters.rb
wioux: It seems to work to wait on initializing the widget until the input is focused, like this:

```js
  $('#filter_action_pages #date_range').on('focus', function() {
    $(this).daterangepicker({
      locale: { format: 'YYYY-MM-DD' }
    });
  });
```

Then listen for the `apply.daterangepicker` event instead of `change` to avoid some weird resulting behavior.

In general though if you have to detect/remove default parameters it's nice to do that in a before_action.

app/models/action_page.rb
```diff
@@ -77,6 +78,20 @@ def action_type
...
+    unless %w(archived victory live draft).include?(status)
+      raise ArgumentError, "unrecognized status #{status}"
+    end
+    case status
+    when "live"
+      where(published: true, archived: false, victory: false)
+    when "draft"
+      where(published: false, archived: false, victory: false)
+    else
+      where(status.to_sym => true)
```
wioux: `to_sym` isn't needed here.

app/views/admin/action_pages/index.html.erb
```diff
@@ -5,7 +5,28 @@
   <%= form_tag admin_action_pages_path, id: "filter_action_pages", method: :get do %>
     <div class="form-item">
       <%= label_tag "Type" %>
-      <%= select_tag :action_type, options_for_select(%w(all call congress_message email petition tweet redirect)) %>
+      <%= select :action_filters, :type, options_for_select(%w(all call congress_message email petition tweet redirect)) %>
+    </div>
+
+    <div class="form-item">
+      <%= label_tag "Date Range" %>
+      <input type="text" class="form-control text-inline" id="date_range" name="date_range"
```
wioux: The `form-control text-inline` classes are from the old stylesheet and aren't needed anymore.

app/controllers/admin/action_pages_controller.rb
```diff
@@ -18,6 +18,11 @@ class Admin::ActionPagesController < Admin::ApplicationController
   allow_collaborators_to :index, :edit
 
   def index
+    @categories = Category.all.order(:title).map { |c| [c.title, c.id] }
+    @authors = User.joins(:action_pages).order(:last_name).map do |a|
+      label = a.name.blank? ? a.email : a.name
+      [label, a.id]
+    end
```
wioux: Formatting view helper arguments in the controller like this is a bit unconventional. It'd be better for User to have an `.authors` scope and `#display_name` method, and feed these into Rails' `options_from_collection_for_select` helper.

app/models/user.rb
```diff
@@ -26,6 +27,8 @@ class User < ActiveRecord::Base
 
   alias :preferences :user_preferences
 
+  scope :authors, ->() { joins(:action_pages) }
```
wioux: Need a `.distinct` call here, otherwise the select will have duplicate options.

app/views/admin/action_pages/index.html.erb
```diff
@@ -5,7 +5,27 @@
...
+    </div>
+
+    <div class="form-item">
+      <%= label_tag "Author" %>
+      <%= collection_select :action_filters, :author, @authors, :id, :display_name, { include_blank: "all" } %>
+    </div>
+
+    <div class="form-item">
+      <%= label_tag "Category" %>
+      <%= collection_select :action_filters, :category, @categories, :id, :title, { include_blank: "all" } %>
```
wioux: Didn't know about `collection_select`, nice! `options_from_collection_for_select` is so ugly..

PR #601, opened 2019-08-07
Only display categories that exist

esoterik:


esoterik:
this was handled elsewhere, closing

PR #602, opened 2019-08-08
Create status enum on action pages

esoterik:
Draft for now -- this is more or less just a database migration and model level changes required to change status into an enum. It has not been extensively tested, so I suspect that there might be some other broken things or improvements to be made elsewhere!

### DIFF
--- a/app/queries/action_page_filters.rb
+++ b/app/queries/action_page_filters.rb
@@ -14,7 +14,9 @@ class ActionPageFilters
 
     filters.each do |f, val|
       next unless valid_query?(f, val)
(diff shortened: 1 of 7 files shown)


app/queries/action_page_filters.rb
```diff
@@ -14,7 +14,9 @@ def run
 
     filters.each do |f, val|
       next unless valid_query?(f, val)
-      @relation = if NAMED_SCOPES.include? f
+      @relation = if ENUM_QUERY == f
+                    relation.send f
```
esoterik: this should send `val`, not the filter name

PR #608, opened 2019-08-12
remove google plus from email shares

### LINKED ISSUE #603: Remove G+ button from thank you email
The thank you email you receive after taking an action still includes the Google+ share button. Can we remove that?
Screen Shot 2019-08-08 at 12 49 32 PM

lschatzkin:
closes #603


app/views/user_mailer/_sharing.html.erb
```diff
@@ -5,9 +5,6 @@
   <a href='<%= twitter_share_url(@actionPage) -%>'>
     <%= image_tag 'share-on-tw-thin.png', alt: 'Share on Twitter' -%>
   </a>
-  <a href='<%= google_share_url(@actionPage) -%>'>
```
esoterik: we can also get rid of this method which is in `app/helpers/action_page_helper.rb`, then good to merge!

PR #619, opened 2019-08-14
497 next page

lschatzkin:
I'm reviving this branch as it seemed to have gotten lost in the shuffle. Updating the styles. Still need to have four pages instead of three, and some js to change the colors.


lschatzkin:
@vbrown608 I've reached a sticking point on this for the "page 4", which is the thank you page.  I'm using jquery to change the progress bar but that page is rendered via an if statement views/tools/_container.html.erb. Is there a way to get that statement to trigger js?

esoterik:
> @vbrown608 I've reached a sticking point on this for the "page 4", which is the thank you page. I'm using jquery to change the progress bar but that page is rendered via an if statement views/tools/_container.html.erb. Is there a way to get that statement to trigger js?

Not sure if this will help, but the thank you page is shown via the various JS files in the `app/assets/javascripts/application/tools` folder -- there's a line to hide the previous page and then show the thank you (`$('#thank-you').show();`). It's code duplication but you could update the bar near there!

lschatzkin:
> > @vbrown608 I've reached a sticking point on this for the "page 4", which is the thank you page. I'm using jquery to change the progress bar but that page is rendered via an if statement views/tools/_container.html.erb. Is there a way to get that statement to trigger js?
> 
> Not sure if this will help, but the thank you page is shown via the various JS files in the `app/assets/javascripts/application/tools` folder -- there's a line to hide the previous page and then show the thank you (`$('#thank-you').show();`). It's code duplication but you could update the bar near there!

Yes, I'm editing the congress_message.js file in that folder... just can't quite get this last bit working.

app/assets/javascripts/application/tools/congress_message.js
```diff
@@ -4,6 +4,43 @@ $(document).on("ready", function() {
...
+    });
+    $(".page-indicator div.page3").css({
+      'background-color': '#BABABA',
+      'color': '#2D2D2D'
+    });
+  });
+
+  $("#tools #congress-message-tool input[type='submit'].test").click(function(){ // show thank you
+
+  });
```
esoterik: oh, sorry, one last thing! this click handler can be deleted now

lschatzkin: Oh, right!

PR #638, opened 2019-08-16
Check if an image is present more safely

esoterik:
Allows actions without images to be viewed -- if there's no image object right now, `#exists?` throws an error.


lschatzkin:
Lovely.

PR #640, opened 2019-08-19
Add analytics calculations to admin action page table

esoterik:


app/views/admin/action_pages/_index.html.erb
```diff
@@ -18,9 +18,9 @@
         <td><%= action_page.action_type %></td>
         <td><strong><%= link_to action_page.title, edit_admin_action_page_path(action_page) %></strong></td>
         <td class="nowrap"><%= action_page.created_at.strftime("%m-%d-%Y") %></td>
-        <td></td>
-        <td></td>
-        <td></td>
+        <td><%= action_page.event_summary[:views] %></td>
```
esoterik: this is a pretty awful N+1 query, but adding an `includes(:events)` to the controller query in `#filter_action_pages` is causing my rails server to lock up, then crash, locally. I'm investigating solutions now.

esoterik: going to add a couple of counter caches!

esoterik:
okay, ready for review! conditional counter caches were implemented with [this gem](https://github.com/magnusvk/counter_culture), which is well-maintained. 

I populated the caches in a db migration, but we can also use a rake task with [something like this](https://github.com/magnusvk/counter_culture#manually-populating-counter-cache-values). The gem's maintainer recommends running a task like that regularly but I figured that might be best in a separate PR because that would involve a new cron job

esoterik:
closing and reopening to merge into master

PR #641, opened 2019-08-21
fixed privacy notice on non-congress form pages

lschatzkin:


esoterik:
i think this is missing some JS for the tooltips, they're still just shown for me:
image

lschatzkin:
> i think this is missing some JS for the tooltips, they're still just shown for me:
> image

Weird! It's CSS only. I'll see if I can replicate.

lschatzkin:
I tried merging the latest merge-redesigns into the branch, but still am not seeing the brokenness.

esoterik:
it looks fine for me now!

PR #666, opened 2019-08-29
652 effector checkbox

### LINKED ISSUE #652: Make EFFector checkbox checkable on congress message actions
The “Yes! I would like to join…” effector checkbox is uncheckable here: https://act.staging.eff.org/action/tell-the-senate-to-restore-full-net-neutrality-protections
Screen Shot 2019-08-28 at 2 53 21 PM

lschatzkin:
EFFector/newsletter signup js is interfering with the the css-styled checkbox, so reverting to non-styled for this one element for now.  Should restyle at some point but already have spent several hours debugging and need to move on.

closes #652

### DIFF
--- a/app/assets/stylesheets/application/_forms.scss
+++ b/app/assets/stylesheets/application/_forms.scss
@@ -132,6 +132,23 @@ input[type=checkbox]:checked + label:after,
   color: #aaa;
 }
 
+// congress forms email-signup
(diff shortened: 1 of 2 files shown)


esoterik:
I'm seeing a bit of an alignment error with the text:

image

esoterik:
also, is that the correct location for this checkbox? seems like it should go at the bottom of the form, but I'm not sure

PR #674, opened 2019-08-30
adding hover state to more action header h3

### LINKED ISSUE #669: Remove hover color on action summary text
Currently, the Action summary text turns gray on hover. Please remove that.
Screen Shot 2019-08-29 at 3 06 01 PM

lschatzkin:
instead of removing hover from summary
closes #669


esoterik:
The hover states can all happen independently, so I'm seeing some unexpected behavior: the image is greyed out when I hover over it, all of the text is greyed when i hover over the action title, and only the summary is greyed when i hover over it. I think it'd be better to at least unify the text hovers so that hovering over either greys all of the text out. What do you think?

lschatzkin:
I see what you mean. I'll change it so they all gray out together.

esoterik:
i'm still seeing the summary hover not greying the title in firefox

lschatzkin:
Oh, I only fix the index page! Fixing home page now.

PR #676, opened 2019-09-03
Fix congress message actions

### LINKED ISSUE #650: Add Error Message to Congress Message Action
For a congress message action, the tool breaks with no error message when an address is misentered or no reps are found
Screen Shot 2019-08-28 at 2 48 26 PM

esoterik:
closes #650

### DIFF
--- a/app/views/congress_messages/_form.html.erb
+++ b/app/views/congress_messages/_form.html.erb
@@ -40,9 +40,9 @@
   <%= hidden_field_tag :bioguide_ids, @message.targets %>
 
   <div class="submit-wrapper">
-    <input type="submit" value="Next >" class="btn action">
+    <input id="to-page-3" type="submit" value="Next >" class="btn action">
     <% if current_user && current_user.privileged_role? %>
-      <input type="submit" value="Test" name="test" class="btn action test">
+      <input id="to-page-3" type="submit" value="Test" name="test" class="btn action test">
     <% end %>
   </div>
   <%= render "tools/loading" -%>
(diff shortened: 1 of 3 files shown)


esoterik:
This also corrects the order of the forms, as it should be Find Reps -> Additional Info -> Customize Message

app/views/congress_messages/_form.html.erb
```diff
@@ -40,9 +40,9 @@
   <%= hidden_field_tag :bioguide_ids, @message.targets %>
 
   <div class="submit-wrapper">
-    <input type="submit" value="Submit" class="btn action">
+    <input id="to-page-3" type="submit" value="Next >" class="btn action">
     <% if current_user && current_user.privileged_role? %>
-      <input type="submit" value="Test" name="test" class="btn action test">
+      <input id="to-page-3" type="submit" value="Test" name="test" class="btn action test">
```
lschatzkin: Can we make these class="eff-button" ? I'm trying to get rid of those bootstrap buttons.

lschatzkin:
Step 3 never appears for me, I just end up at the thank you with progress numbers in the wrong sequence.
Screenshot from 2019-09-06 13-23-31

esoterik:
also closes #689!

esoterik:
ready for re-review :smiley_cat:

lschatzkin:
Hmm, I'm still not having the same issue i.e. customize messages page is not appearing, I just go straight to thank you page with the incorrect progress numbers.

lschatzkin:
YES! This is working.
One tiny 
message-input
problem: there seems to be an errant message input field in the Common Fields.

esoterik:
unfortunately that's a separate congress forms bug -- it made the assumption that the custom message would be written first. mind if i open up a separate ticket for that?

lschatzkin:
Sure, I will go ahead and merge this and you can work on that issue. If it's too bothersome, we can just hide it with css.

PR #682, opened 2019-09-04
681 register

### LINKED ISSUE #681: Register/sign up
On the new action center, we still don't explain what "Register" means. This could be solved by adding a line of brief text directly under "Sign Up" (And, actually, changing "Sign Up" to "Register" to match the term in the drop down. 

Text: Registering for the Action Center allows you to keep track of any EFF actions you have taken, save information like your address for convenience in taking actions later, and manage your subscriptions to EFF newsletters. Within your account you may choose not to keep track of your actions, not supply any information, or subscribe to newsletters, if that is your preference.  All information is subject to EFF's privacy policy, and you may delete your account at any time.

lschatzkin:
closes #681

### DIFF
--- a/app/assets/stylesheets/application/_pages.scss
+++ b/app/assets/stylesheets/application/_pages.scss
@@ -15,7 +15,14 @@
   }
 }
 
-
+.registrations-new {
+  .row div p{
+    position: relative;
+  }
+  span.register-info {
+    @include eff-tooltip;
+  }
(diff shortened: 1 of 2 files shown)


esoterik:
could the tooltip trigger be on its own line or have a bit more space between it and the text? When I look at it right now I'm not sure what it's explaining.

Issue #647, opened 2019-08-28
Enable Petition Action


krcarlson:
Filling out petitions doesn’t do anything.
<img width="663" alt="Screen Shot 2019-08-28 at 2 28 26 PM" src="https://user-images.githubusercontent.com/8976353/63894679-17a3b380-c9a2-11e9-8bf6-b456dfb880de.png">

PR #684, opened 2019-09-05
Fix petition issues

### LINKED ISSUE #648: Fix 500 error when Petition Action goal = 0
Small bug:
Pages return a 500 error when the petition has goal=0 
https://act.staging.eff.org/action/keep-safe-harbors-safe-ed9c6cc0-8f55-4945-a760-659c7e145536

esoterik:
closes #648

### DIFF
--- a/app/views/tools/_petition.html.erb
+++ b/app/views/tools/_petition.html.erb
@@ -10,7 +10,6 @@
     </div>
 
     <div class="tool-body">
-      <form>
       <%- if @petition.enable_affiliations -%>
         <div class='small tool-notice'>
           <%= t :name_and_affiliations_published -%>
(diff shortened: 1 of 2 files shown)


esoterik:
also closes #647 !

lschatzkin:
LGTM!

app/views/tools/_petition.html.erb
```diff
@@ -38,8 +37,7 @@
           <%= render 'tools/thank_you' -%>
         </div>
         <div class='unsigned'>
-          <%= form_for @signature, url: '/tools/petition',
-            html: {class: 'petition-tool'} do |f| -%>
+          <%= form_for @signature, url: tools_petition_path, html: {class: 'petition-tool'} do |f| -%>
```
lschatzkin: Just so I know, where is the tools_petition_path get set?

esoterik: all of these path helpers are created in [this file](https://github.com/EFForg/action-center-platform/blob/master/config/routes.rb#L12) (the line I linked to makes that one!)

`config/routes.rb` does more than that though -- it controls what URLs go to what controller methods. I often reference [this guide](https://guides.rubyonrails.org/routing.html) when I'm making routing changes.

lschatzkin: Cool, thank you!

PR #696, opened 2019-09-09
"Signed" now visible, plus a couple of style changes

### LINKED ISSUE #692: Change color of "Signed" text on petition action
Currently the word "Signed" is not legible/accessible on the petition action
Screen Shot 2019-09-09 at 11 34 01 AM

lschatzkin:
closes #692

### DIFF
--- a/app/assets/stylesheets/action_page.scss
+++ b/app/assets/stylesheets/action_page.scss
@@ -634,17 +634,16 @@ h3.header,
       font-family: serif;
     }
     .signed {
-      color: $gray-light;
+      color: $charcoal;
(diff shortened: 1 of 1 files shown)


esoterik:
I just noticed that the font family isn't set in the petition letter -- it's showing in my browser's default serif font. Is it supposed to be Monserrat serif?

lschatzkin:
Yes it is, I'll check that.

lschatzkin:
Actually, I think we should leave this as is. I think the plain Times font looks better in indicated that it's a letter.

PR #704, opened 2019-09-09
updating spacing for category headers

### LINKED ISSUE #700: Reduce header space on All Campaigns page
Reduce the space between the two heading styles by half. Where no action type header is included, make the heading have the same distance from the image as the action type header. See attached png.
action-all-spacing

lschatzkin:
closes #700

### DIFF
--- a/app/assets/stylesheets/application/_base.scss
+++ b/app/assets/stylesheets/application/_base.scss
@@ -242,6 +242,7 @@ code {
 h5.category {
   color: $gray;
   font-size: $base-font-size * 0.9;
+  padding-bottom: 0;
   padding-top: 1rem;
 }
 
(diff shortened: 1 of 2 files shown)


esoterik:
I'm seeing a computed 16px spacing between the category and the image, while there's just 8px between the title and the image. Maybe it'd be better to set this space on the `.text` div and put the category in there to ensure that the spacing is the same?

lschatzkin:
> I'm seeing a computed 16px spacing between the category and the image, while there's just 8px between the title and the image. Maybe it'd be better to set this space on the `.text` div and put the category in there to ensure that the spacing is the same?

I tried this out, but one still ends up with different/tweaky passing because the category and the title are different font sizes so visually it has to be adjusted.

PR #706, opened 2019-09-09
remove flicker from footer logo

### LINKED ISSUE #701: Remove animation from EFF logo in footer
It looks like the EFF logo in the footer has some sort of animation on it (it flickers on hover). It should just turn white on hover.

lschatzkin:
closes #701

### DIFF
--- a/app/assets/stylesheets/application/_footer.scss
+++ b/app/assets/stylesheets/application/_footer.scss
@@ -50,13 +50,14 @@
     .logo {
       width: 28rem;
       a {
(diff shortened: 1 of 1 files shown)


esoterik:
hmm, I still see a flicker -- it only happens on the first time I hover over the logo after the page load. I added a similar hover in certbot by using the white logo everywhere and adjusting the opacity on hover, could that work here?

lschatzkin:
I was thinking of trying that yesterday, but then it seemed fixed! I'll try that now.

Issue #693, opened 2019-09-09
Admin: Preview button doesn't do anything


krcarlson:
Clicking Preview in the Action Center admin create new action flow doesn't do anything.

krcarlson:
It seems this is still an issue https://redmine.eff.org/issues/22930

PR #710, opened 2019-09-10
Fix previews

esoterik:
sorta for #693 -- previews were broken after adding related content. This doesn't fix previews for new actions (the functionality was originally written only for editing actions, and we're leaving that as-is for now)


lschatzkin:
Preview still isn't working for me.

lschatzkin:
nm, got it to work.

PR #716, opened 2019-09-11
Allow users to contact select congress members

esoterik:
WIP

### DIFF
--- a/app/assets/javascripts/application/tools/congress_message.js
+++ b/app/assets/javascripts/application/tools/congress_message.js
@@ -29,6 +29,10 @@ $(document).on("ready", function() {
     $("#customize-message").show();
     update_tabs(2, 3);
   });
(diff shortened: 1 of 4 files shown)


lschatzkin:
Okay, I'm not sure what I should be seeing so far, but this is what I see.  The congress people's names show up, the checkboxes are currently hidden (as expected). When I show the checkboxes I don't see them having any effect. I'm not sure if this is the part you'd like me to do?
Screenshot from 2019-09-11 08-55-40
Screenshot from 2019-09-11 08-55-57

esoterik:
I'll handle the JS that shows & hides the subsequent forms when the checkboxes change. It works on the backend right now. I was hoping for some help with the styling of the checkboxes!

esoterik:
the JS has been added!

lschatzkin:
I'm going to add myself a ticket to try to improve the look of these checkbox in the future, but for now they are minimally styled.

PR #726, opened 2019-09-11
702 nav menu

lschatzkin:
The problem here is that the nav drop down is a modal, so that the user can click anywhere to close. This means it is a child of the entire screen, so aligning it with a the hamburger at all widths isn't really possible without js.  I've added a few things to make it align a little better.


esoterik:
oh i just merged into master -- can you switch the branch?

lschatzkin:
to master?

esoterik:
yup!

lschatzkin:
done!

PR #736, opened 2019-09-12
Match text sizes on call actions

### LINKED ISSUE #730: Match text sizes on call actions
Can we make text under "Before You Call" the same size as text in the "What To Say" box? Looks like the Before You Call text is a bit bigger right now. 
 
Screen Shot 2019-09-11 at 3 40 32 PM

lschatzkin:
closes #730


esoterik:
can these match on mobile as well?

lschatzkin:
> can these match on mobile as well?

Oh crap, sorry I missed that. Here it is.

PR #743, opened 2019-09-18
Fix congress message actions that target specific members

esoterik:


app/assets/javascripts/application/tools/congress_message.js
```diff
@@ -1,6 +1,23 @@
...
   if ($("#congress-message-tool").length){  // showing paging at top of form
     $(".page-indicator").css("display", "grid");
+    if ($(".page-indicator").children().length == 3) {
+      $(".page-indicator").css({
+        "grid-template-columns": "33% 33% 33%"
+      });
+    }
+  }
+
+  if ($(".load-target-members").length && $(".load-target-members").is(":visible")) {
```
wioux: The `.length` check here looks redundant -- if the element is missing, it shouldn't be visible.

PR #749, opened 2019-09-24
changing logo to monogram on mobile

lschatzkin:


esoterik:
should it still say "Action Center" next to the logo on mobile?

lschatzkin:
I think not but let's check... @krcarlson  can you ask Hugh? If so, we would need an image from him to use.

lschatzkin:
added new mobile log from Hugh

PR #750, opened 2019-09-24
topic tooltip added

lschatzkin:


esoterik:
I think we can get the topics tooltip to be closer to the dropdown by adding the tooltip HTML to the `congress_forms_field` method in `app/helpers/congress_message_helper`. something like the following should work:
```ruby
elsif field.value == "$TOPIC"
  "#{select_tag [same as what's there for elsif field.is_select?]}\n"\
  "<html-for-topics-tooltip>"
```
let me know if that makes sense!

lschatzkin:
> I think we can get the topics tooltip to be closer to the dropdown by adding the tooltip HTML to the `congress_forms_field` method in `app/helpers/congress_message_helper`. something like the following should work:
> 
> ```ruby
> elsif field.value == "$TOPIC"
>   "#{select_tag [same as what's there for elsif field.is_select?]}\n"\
>   "<html-for-topics-tooltip>"
> ```
> 
> let me know if that makes sense!

Very cool! I think there is also a design concern that having the tooltip next to the dropdown means it's outside the margin. @krcarlson can you check on that?

krcarlson:
Yeah, Hugh wanted to try the tool tip next to the representative's name (right above the drop down) so it didn't run outside the margin. 

Screen Shot 2019-09-24 at 10 06 33 AM

lschatzkin:
Okay, so then I think I am done.

PR #755, opened 2019-10-10
Re-add petitions with targets to admin interface

esoterik:
WIP, but the functionality is there. 

TODOs:
- [x] Try make the edit targets page load faster, though it's less bad than it was
- [ ] Figure out what to do when an institution is deleted (e.g. delete associations), maybe block deleting institutions that have associated signatures
- [x] Styling tweaks to institutions table (notably, make button links styled properly)
- [x] Figure out why firefox gives a corrupted content error after successful CSV upload


db/migrate/20191002224831_add_category_to_institutions.rb
```diff
@@ -0,0 +1,5 @@
+class AddCategoryToInstitutions < ActiveRecord::Migration[5.0]
+  def change
+    add_column :institutions, :category, :integer, null: false, default: 0
```
esoterik: This should maybe not have a default value and instead just update all existing institutions to be universities, thoughts?

app/views/admin/institutions/edit_targets.html.erb
```diff
@@ -0,0 +1,75 @@
...
+      </div>
+
+      <% @grouped_institutions.each do |category, institutions| %>
+        <div id="<%= category %>-institutions" class="institutions-by-category hidden">
+          <div class="alert">All <%= category.humanize.pluralize.downcase %> have been added!</div>
+          <%= button_tag "Edit included #{category.humanize.pluralize.downcase}", type: "button",
+            class: "btn btn-default category-show", id: "show-all-#{category}" %>
+          <%= button_tag "Reset targets", type: "button", class: "btn btn-default", id: "reset-institutions" %>
+          <div class="institutions-list">
+            <%= f.collection_check_boxes :institution_ids, institutions, :id, :name,
```
esoterik: It might be nice to instead have a select2 box that allows individual institutions to be removed, like the way editing works.

app/controllers/admin/institutions_controller.rb
```diff
@@ -1,64 +1,69 @@
...
+    redirect_to "index", notice: "Target deleted"
   end
 
-  # DELETE /admin/action_pages/:action_page_id/institutions/
-  def destroy_all
-    @actionPage.institutions.delete_all
-    respond_to do |format|
-      format.html { redirect_to [:admin, @actionPage, Institution] }
-    end
+  def edit_targets
```
esoterik: this and `update_targets` might actually fit better in a new controller (`Admin::ActionInstitutionsController`) because it deals with creating and removing those join models.

app/controllers/admin/institutions_controller.rb
```diff
@@ -1,64 +1,69 @@
...
+    @not_included_institutions = @institutions - @action_institutions
+    @grouped_institutions = @institutions.group_by(&:category)
+    5.times { @actionPage.affiliation_types.build }
+  end
+
+  def update_targets
+    @actionPage.update(affiliation_types_params)
+    ActionInstitution.update_action_page(action_page: @actionPage,
+                                         add: institutions_diff_params[:add_ids],
+                                         remove: institutions_diff_params[:remove_ids])
```
esoterik: possibly should do this in the background like importing institutions, it takes a while

app/views/admin/action_pages/_petition_fields.html.erb
```diff
@@ -8,4 +8,17 @@
...
+      Add petition targets <span class="hint">(ex. universities or cities)</span>
+    <% end %>
+    <%= sf.check_box :enable_affiliations %>
+  </div>
 <% end %>
+
+<div id="affiliations-enabled" <%= 'class="hidden"' unless @actionPage.petition.enable_affiliations %>>
+  <%= link_to "Edit petition targets and affiliations",
+    edit_targets_admin_action_page_institutions_path(@actionPage),
+    target: "_blank" %>
```
esoterik: this link will only work on persisted action pages, should probably surface this link better (maybe in the sidebar on edit, only when affiliations are enabled?)

esoterik:
Updated; only TODOs left are

- [ ] Institution views styling tweaks
- [ ] figure out what to do when institutions are deleted

Another potential issue: when new institutions are uploaded, if one of the name already exists, then it will just update the existing institution. This is only a problem if the category is different -- it will update the category. It won't touch associations, but this has the potential to cause some issues.

db/migrate/20191002224831_add_category_to_institutions.rb
```diff
@@ -0,0 +1,7 @@
+class AddCategoryToInstitutions < ActiveRecord::Migration[5.0]
+  def change
+    add_column :institutions, :category, :string
+    Institution.update_all(category: "university")
```
esoterik: it looks like there's some states and cities currently in the database (or at least in my kind of old local copy), so some individual edits might be necessary

esoterik:
rebased and cleaned up commit history + added Laura's styling fixes for the user-facing side

app/controllers/admin/action_pages_controller.rb
```diff
@@ -51,10 +58,13 @@ def create
   def edit
     @actionPage.petition ||= Petition.new
     @actionPage.tweet ||= Tweet.new
-
     @actionPage.call_campaign ||= CallCampaign.new
     @actionPage.email_campaign ||= EmailCampaign.new
     @actionPage.congress_message_campaign ||= CongressMessageCampaign.new
+    5.times { @actionPage.affiliation_types.build }
```
esoterik: we should bump this up to 10

app/controllers/admin/action_pages_controller.rb
```diff
@@ -36,12 +38,17 @@ def new
...
     @actionPage.congress_message_campaign = CongressMessageCampaign.new
     @actionPage.email_text = Rails.application.config.action_pages_email_text
+    5.times { @actionPage.affiliation_types.build }
   end
 
   def create
     @actionPage = ActionPage.new(action_page_params.merge(author: current_user))
 
     if @actionPage.save
+      if institutions_params[:category]
```
wioux: This condition should trigger `ActionInstitution.delay.add` in the update action as well. That said, it looks like it may make more sense to add a `institution_category` column to action pages, and use that as the foreign key for institutions? Delaying the import here also seems to create a race condition when the page reloads, due to this line

```ruby
@target_category = @actionPage.institutions.first.category
```

app/controllers/admin/action_pages_controller.rb
```diff
@@ -66,6 +76,10 @@ def update
     @actionPage.og_image         = nil if params[:destroy_og_image]
 
     @actionPage.update_attributes(action_page_params)
+    if institutions_params[:reset] && institutions_params[:reset] == "1"
+      ActionInstitution.add(action_page: @actionPage,
```
wioux: This should also trigger when `institutions_params[:category]` is present, otherwise the branch looks ready!

esoterik: done!

wioux: The situation I was thinking of here is when affiliations are enabled for the first time on an already-existing petition. In that case, the category param is present but not the reset param, so the `ActionInstitution.add` call here is still missed and the page hits an error when it reloads, on 

```ruby
@institution_category = @institutions.first.category
```

esoterik: OH! that makes sense, just changed it. It only triggers when there's no institutions on the action.

PR #772, opened 2019-11-04
Enable partners admin + add logos

esoterik:
~~The partner logos div is currently set to hidden; I'll wait on styling it until we have some logo images that we want to display.~~


app/views/action_page/_partner_logos.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if @actionPage.partners.present? %>
+  <div class="partner-logos">
+    <% @actionPage.partners.each do |partner| %>
+      <%= image_tag partner.logo %>
```
esoterik: needs an `if partner.logo.present?`

esoterik:
Here's what the logo display looks like at different screen sizes, with different numbers of partners:

desktop:
action center logos desktop

tablet:
action center logos ipad

mobile:
action center logos mobile 1

esoterik:
this needs a rebase, will do today

db/schema.rb
```diff
@@ -145,6 +143,19 @@
     t.datetime "updated_at"
   end
 
+  create_table "ckeditor_assets", force: :cascade do |t|
```
esoterik: this got mistakenly added from my local db, fixing now

app/views/admin/action_pages/edit_partners.html.erb
```diff
@@ -0,0 +1,29 @@
+<% content_for(:nav) { render "action_page_nav" } %>
+
+<h2>Partners for <%= @actionPage.title %></h2>
+
+<p> Add partner logos and email signups to this action. </p>
+
+<%= form_for([:admin, @actionPage]) do |f| %>
```
wioux: Just noticed while testing this that #755 broke the update action for forms like this -- the call to `institutions_params` assumes that the request is made from the main form

```ruby
  def update
  ...
      @actionPage.update_attributes(action_page_params)
      if (institutions_params[:reset] && institutions_params[:reset] == "1") ||
  ...
  end

  def institutions_params
    params.require(:institutions).permit(%i(category reset))
  end
```
```

File: app/helpers/action_page_helper.rb [hunk 4]
```diff
@@ -77,7 +77,9 @@ def parse_email_text(options = {})
   end
 
   def visible_partners
-    featured = @actionPage.partners.where(code: params[:partner])
-    featured.empty? ? @actionPage.partners : featured
+    mailings_enabled = @actionPage.partners.includes(:partnerships)
+                                  .where(partnerships: { enable_mailings: true })
+    return mailings_enabled if params[:partner].empty?
```
wioux: I get an error here -- should be `params[:partner].blank?` instead?

PR #775, opened 2019-11-08
Count when congress messages are customized

esoterik:


app/controllers/admin/action_pages_controller.rb
```diff
@@ -127,7 +127,13 @@ def duplicate
 
   def events
     respond_to do |format|
-      format.html
+      format.html do
+        if @actionPage.enable_congress_message?
+          @total = @actionPage.events.count
```
esoterik: need to exclude views

esoterik: should also start counting after a certain date near the launch of this feature

app/controllers/congress_messages_controller.rb
```diff
@@ -100,8 +100,10 @@ def subscribe_user
   end
 
   def track_action
+    customized_message = params[:message] == @campaign.message
```
wioux: Looks like this should be `!=`? Otherwise this looks good to me. :+1:

app/controllers/admin/action_pages_controller.rb
```diff
@@ -129,7 +129,9 @@ def events
     respond_to do |format|
       format.html do
         if @actionPage.enable_congress_message?
+          launch_date = Time.zone.parse("Nov 19 2019")
           action_events = @actionPage.events.where(name: 'Action')
+                                            .where("time > ?", launch_date)
```
wioux: I think you could avoid the magic date here by only looking at events where the new property is present -- for example:

```ruby
where("json_extract_path(properties, 'customizedMessage') is not null")
```

PR #777, opened 2019-11-13
Fix #746 "Save removes paragraphs"

### LINKED ISSUE #746: Admin: Save removes paragraphs
When admin create a new action (in this case, a congress message) and click "Save" on the last step  (Thank you page), the paragraphs/line breaks are removed from the Summary and Description text when you view it. It seems like you can go back and edit the action to add back in the paragraphs, but it's a pain.
Screen Shot 2019-09-19 at 10 05 57 AM

wioux:
Epiceditor doesn't correctly export line breaks into the textarea when the iframe element is invisible. The textarea is automatically updated [every 100ms](https://github.com/OscarGodson/EpicEditor/blob/15802f057222d12ce4a0185594092b932a7127b2/src/editor.js#L1158-L1163) (even when autoSave is set to false `:\`), so this means as the user moves through tabs while editing an action page, content in the editors they're navigating away from gets overwritten with an incorrect value.

This branch provides a quick fix for the issue by linking the editor's `_canSave` attribute (which is internal state?) to it's visibility.

### DIFF
--- a/app/assets/javascripts/admin/editor.js.erb
+++ b/app/assets/javascripts/admin/editor.js.erb
@@ -27,5 +27,6 @@ function reflowEpicEditor() {
       initEpicEditor(this);
 
     $(this).data('editor').reflow();
+    $(this).data('editor')._canSave = true;
   });
 }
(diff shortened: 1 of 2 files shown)


esoterik:
oh great!! :tada:

PR #784, opened 2019-11-26
Update signatures view + allow signatures to be exported

esoterik:


app/controllers/admin/petitions_controller.rb
```diff
@@ -20,6 +21,11 @@ def presentable_csv
     send_data @petition.to_presentable_csv, filename: filename
   end
 
+  def affiliation_csv
+    filename = sanitize_filename("#{@petition}.csv")
+    send_data @petition.to_affiliation_csv, filename: filename
+  end
+
   def destroy_signatures
```
wioux: Looks like the destroy_signatures action got broken by the redesign --

```ruby
undefined method `admin_petition_path' 
```

app/models/petition.rb
```diff
@@ -53,12 +53,13 @@ def to_presentable_csv(options = {})
     end
   end
 
-  def to_affiliation_csv(institution, options = {})
+  def to_affiliation_csv(options = {})
```
wioux: The call to this in `app/controllers/action_page_controller.rb` still uses the old method signature:

https://github.com/EFForg/action-center-platform/blob/3556b0c0bd36b40783dc3fdd392d27649f2c34ef/app/controllers/action_page_controller.rb#L60-L65

wioux:
Looks good to me, just needs conflicts fixed. :+1:

PR #787, opened 2019-12-02
Add view, action counts to admin action page table

esoterik:


app/views/admin/action_pages/_index.html.erb
```diff
@@ -18,9 +18,9 @@
         <td><%= action_page.action_type %></td>
         <td><strong><%= link_to action_page.title, edit_admin_action_page_path(action_page) %></strong></td>
         <td class="nowrap"><%= action_page.created_at.strftime("%m-%d-%Y") %></td>
-        <td></td>
-        <td></td>
-        <td></td>
+        <td><%= action_page.view_count %></td>
+        <td><%= action_page.action_count %></td>
+        <td><%= number_to_percentage(action_page.actions_taken_percent, precision: 2) %></td>
```
wioux: Can we use `precision: 0`? I think it makes the table a little easier to read.

wioux:
Do you have any ideas about how we wind up with `action_count` greater than `view_count` on some pages? My only thought was that some add/tracker blockers will block the request to visit.gif, especially if the action is embedded on another domain -- but I'm wondering if there are other scenarios too.

esoterik:
I'm not sure. I had thought there might be some inconsistencies in how views are recorded -- when I was working on this PR, I was basing view counts from the scope:
```ruby
scope :views,      -> { where(name: "View") }
```
but in the method that calculates the counts for an individual action, there's this line
```ruby
action = "view" if action == "embedded_view" || action == "show"
```
But I took a look and all of those events still have the name "View". From my old local db, it seems like the vast majority of the actions where this is happening are petitions, though there's a couple of email actions and one tweet action as well. That's the only trend I could find.

PR #794, opened 2019-12-12
Improve analytics table for individual actions

esoterik:
WIP, need to:

- [x] Make the sidebar sharable between the table, chart, and site views
- [x] Make specific date entry work
- [ ] Review calculation methods in event model to see if we can get rid of any
- [ ] Figure out if we're getting congress form fill data properly
- [x] Add custom date formatting to event counts? (Possibly needed for the above, would be nice to not show a year if we're not looking at more than a year's worth of time)
- [x] Tweak sidebar styling


esoterik:
I think this is ready for review. It would be nice to clean up the calculation / query methods on events a bit, but it's not 100% necessary.

wioux:
Works nicely! :+1:

PR #799, opened 2020-01-10
move embed.css to public/

lschatzkin:


wioux:
This branch drops the file into public, making it available at `/embed.css` -- but the stylesheet link points to `/stylesheets/embed.css`.

wioux:
Between this and #796 it seems we're just moving the file around with guesswork? Ideally we would leave it in the app/assets/stylesheets directory -- it just needs to be added to the list of precompiled assets [here](https://github.com/EFForg/action-center-platform/blob/master/config/initializers/assets.rb#L4).

esoterik:
I suggested moving it into public to avoid rails adding a hash to the filename, making it hard to link to from other sites, but maybe I misunderstood what was happening with the stylesheet

PR #819, opened 2020-03-02
Appease rubocop

AndyDangerous:
Fix the problems that upset rubocop. In this case, they were all
double/single quotes. e.g.:

```
spec/support/feature_helpers.rb:22:27: C: [Corrected]
Style/StringLiterals: Prefer double-quoted strings unless you need
single quotes to avoid extra backslashes for escaping.
    find(locator).sibling('.select2-container').click
```

Feel free to ignore


AndyDangerous:
updating

PR #827, opened 2020-03-24
Add notice about message customization to congress actions

wioux:
While putting this together I noticed that the margins and paddings inside the tool container are a little tricky to work with. I intend to circle back on that but am opening this PR now so that activism has this notice available for an upcoming campaign.

Two new shades of gray were in the mockup and are added here directly to the stylesheet. I considered adding them to _variables.scss but decided against it. IMO it's not worth it to try and name these colors, as we'll only end up with something like

```scss
$gray-especially-light: #edf1f4;
$gray-lighter: #e5e5e5;
$gray-light: #c4c4c4;
$gray: #BABABA;
$gray-dark: #999999;
$gray-darker: #949494;
$charcoal: #2D2D2D;
// etc
```

### DIFF
--- a/app/assets/stylesheets/action_page.scss
+++ b/app/assets/stylesheets/action_page.scss
@@ -1254,3 +1254,84 @@ html.js #affiliations {
     }
   }
 }
+
+#congress-message-create {
+  position: relative;
+
+  #customize-message .notice {
+    overflow: hidden;
+    max-height: 25em;
+    transition: max-height 250ms linear, padding 250ms ease-in;
+
+    position: absolute;
+    bottom: 0;
+    padding: 1rem 2.5rem;
+    margin: 0 -.8rem -1rem;
+
+    @media (min-width: $md) {
+      margin-left: -2.3rem;
+    }
+
+    background-color: #e5e5e5;
+
+    font-size: .8em;
+
+    &.down {
+      padding-top: 0em;
+      padding-bottom: 0em;
+      max-height: 0em;
+    }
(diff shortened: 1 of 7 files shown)


lschatzkin:
Are we sure that adding the extra grays was intentional? It seems strange that there are seven shades of gray. This isn't a racy novel, after all.

app/assets/stylesheets/action_page.scss
```diff
@@ -1254,3 +1254,84 @@ html.js #affiliations {
...
+    display: flex;
+  }
+
+  #customize-message .notice input {
+    flex: 1;
+    padding: 0.4em;
+    margin: 0;
+    border: 0;
+
+    font-size: 1em;
```
lschatzkin: I see that you are using a combination of ems and rems. We normally try to stick to rems.

wioux: I find ems very useful. Here I need the font size inside the button input to be the same as it is in the parent container. `1em` seems to be the simplest way to do that. Elsewhere I've used `1.2em` to make the `<strong>` font slightly larger than the surrounding text, which is more flexible than hardcoding the container at `0.8rem` and then using `0.96rem` for the strong font size.

Is there a downside to using `em` in cases like this?

lschatzkin: Just a matter of consistency, but I sometimes use pixels too in the way you are using ems. Perhaps we could have a group discussion on this at a team meeting.

PR #834, opened 2020-05-29
Normalize email address for rate limiting

mfb:
Devise allows email addresses to have bounding whitespace, trailing null bytes, or uppercase, normalizing it via strip and downcase.  So our rate limiting should do the same.


wioux:
Looks good to me. The failing travis check seems to be unrelated.

PR #838, opened 2020-06-03
Validate congress message emails

alexymik:
Validate email using regex before creating subscribing to newsletter or progressing in form.


wioux:
Looks good, ty!

PR #859, opened 2021-02-04
Replace poltergeist with apparition; fix failing tests

esoterik:
Travis config is a WIP, opening as a draft for now


esoterik:
given that [travis-ci.org](https://travis-ci.community/t/builds-hang-in-queued-state/10250) isn't running at full capacity, I'm not super concerned about the push build failing, but I'll look into it a little

PR #869, opened 2021-03-16
Fix signature cleanup task; add validations to prevent duplicates

esoterik:
It turns out that this task has been broken for a long time because signatures/subscriptions ids weren't included in the group clause. This is maybe breaking staging at the moment.

### DIFF
--- a/lib/tasks/signatures.rake
+++ b/lib/tasks/signatures.rake
@@ -19,16 +19,20 @@ namespace :signatures do
 
   desc "Deduplicate signatures from petitions and subscriptions from partners"
   task deduplicate: :environment do
-    sig_dups = Signature.group(:petition_id, :email).having("count(*) > 1")
-    sig_dups.pluck(:petition_id, :email).each do |petition_id, email|
-      ids = Signature.where(petition_id: petition_id, email: email).order(:created_at).ids
-      Signature.where(id: ids[1..-1]).delete_all
+    sig_dups = Signature.group(:email, :petition_id).count.map do |data, count|
+      data if count > 1
+    end.compact
+    sig_dups.each do |email, petition_id|
+      Signature.where(petition_id: petition_id, email: email)
+        .order(:created_at).drop(1).map(&:delete)
     end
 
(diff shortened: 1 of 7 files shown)


esoterik:
Notably the new task is a lot less efficient so this should be deployed during off hours.

lib/tasks/signatures.rake
```diff
@@ -19,16 +19,20 @@ namespace :signatures do
...
+      data if count > 1
+    end.compact
+    sig_dups.each do |email, petition_id|
+      Signature.where(petition_id: petition_id, email: email)
+        .order(:created_at).drop(1).map(&:delete)
     end
 
-    sub_dups = Subscription.group(:partner_id, :email).having("count(*) > 1")
-    sub_dups.pluck(:partner_id, :email).each do |partner_id, email|
+    sub_dups = Subscription.group(:email, :partner_id).count.map do |data, count|
```
wtcruft, line 30: I'm not too familiar with this codebase, so I'm curious -- why the switch for how the duplicates are built?

esoterik: The `count(*)` was causing postgres errors because the group statement didn't include subscription ids, and I couldn't find a better way to have the database do the query selection without going deep into SQL. Doing it in ruby is much less efficient but it works now!

wtcruft: Got it, and is removing the task from the crontab the way of enforcing that this task will only be run manually/after hours?

esoterik: yup!

esoterik: yup!

esoterik: Ah, I added model validations to prevent duplicates from getting saved in the first place, so we should only need to run it once after we deploy this!

wtcruft: oh, hooray! well +1 then :) note: I haven't run the action center myself since my sandbox is not completely set up, but codewise lgtm

PR #908, opened 2024-04-04
Fix failing tests in tools_controller_spec.rb

jparr:


spec/controllers/tools_controller_spec.rb
```diff
@@ -93,7 +91,6 @@
       Rails.application.secrets.google_civic_api_key = "test-key-for-civic-api"
 
       stub_request(:get, "http://civic.example.com/?address=%20&includeOffices=true&key=test-key-for-civic-api&levels=administrativeArea1&roles=legislatorUpperBody")
-        .with(headers: request_headers)
```
jparr, line 96: I dont think we need to match on header specific values. This would be valuable if the controller action had different execution paths based on header values. It broke on my machine because of different user agents.

esoterik (approved):

PR #909, opened 2024-04-04
fix congress message spec

jparr:
added a helper for better delayed job error reporting

### DIFF
--- a/app/models/congress_message.rb
+++ b/app/models/congress_message.rb
@@ -69,8 +69,7 @@ class CongressMessage
     return unless valid?
 
     @forms.each do |f|
-      f.delay.fill(attributes_for(f.bioguide_id), campaign.campaign_tag,
-                   test: test)
+      f.delay.fill(attributes_for(f.bioguide_id), campaign.campaign_tag, test)
     end
   end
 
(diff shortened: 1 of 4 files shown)


app/models/congress_message.rb
```diff
@@ -69,8 +69,7 @@ def background_submit(test: false)
     return unless valid?
 
     @forms.each do |f|
-      f.delay.fill(attributes_for(f.bioguide_id), campaign.campaign_tag,
-                   test: test)
+      f.delay.fill(attributes_for(f.bioguide_id), campaign.campaign_tag, test)
```
jparr, line 72: Looks like rubocop 'fixed' this here: https://github.com/EFForg/action-center-platform/commit/9ecfc7d2fcd031ca70c546acc23c36a60ab355b6

spec/spec_helper.rb
```diff
@@ -30,6 +30,12 @@
     mocks.verify_partial_doubles = true
   end
 
+  config.around(:each, :run_delayed_jobs) do |example|
```
jparr, line 33: added this nice helper. it runs the jobs synchronously so you get error reporting inline with the test/logging output

app/models/congress_message.rb
jparr: I need to add something so rubocop doesnt make the same mistake again

jparr: I got rubocop running and it did not try to fix this again so I think its good to go!

esoterik (approved):

PR #910, opened 2024-04-05
fix feature specs

jparr:
This change to database cleaner config fixed a failing testing in features/action_pages/call_action_spec for me.

We should consider removing it as DatabaseCleaner is not usually necessary in new rails versions.

### DIFF
--- a/spec/features/admin/action_creation_spec.rb
+++ b/spec/features/admin/action_creation_spec.rb
@@ -14,9 +14,7 @@ RSpec.describe "Admin action page creation", type: :feature, js: true do
     fill_in "Message", with: "A message"
     next_section
 
-    # skip banner selection
-    next_section
-
+    skip_banner_selection
     fill_in_social_media
 
     tempermental do
@@ -38,9 +36,7 @@ RSpec.describe "Admin action page creation", type: :feature, js: true do
     fill_in "Goal", with: 1000
     next_section
 
-    # skip banner selection
-    next_section
-
+    skip_banner_selection
     fill_in_social_media
 
     tempermental do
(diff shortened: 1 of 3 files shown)


spec/rails_helper.rb
```diff
@@ -31,6 +31,7 @@
 Capybara.server = :puma
 Capybara.javascript_driver = :selenium_chrome_headless
 Capybara.enable_aria_label = true
+Capybara.disable_animation = true
```
jparr, line 34: This should make the weird timing problems in admin/action_creation_spec go away. I didnt remove everything in there but it should remove the need for sleeps.

spec/features/admin/action_creation_spec.rb
```diff
@@ -152,15 +138,21 @@ def fill_in_basic_info(title:, summary:, description:)
...
+    select(category.title, from: "action_page_category_id") #we should use the label to find the select but the label here is not correctly pointing to the select
   end
 
   def next_section
     click_on "Next"
     sleep 0.05
   end
 
+  def skip_banner_selection
+    expect(page).to have_selector("#images", visible: true, wait: 5)
```
jparr, line 150: Doing an expect like this is a better way to make sure a page is settled before calling actions that will error if an element isnt visible. Even though I'm waiting for potentially 5 seconds it will automatically continue as soon as it resolves true. You also know that the element isnt available which is a more useful error than what shows if it cant fill in an input.

esoterik: oh, that's a great pattern!!

esoterik (approved):

PR #911, opened 2024-04-05
Fixing deprecation warnings

jparr:


Gemfile
```diff
@@ -89,7 +89,7 @@ gem "will_paginate", "~> 3.0"
 gem "xmlrpc", "~> 0.3"
 
 # For creating many records, quickly
-gem "fast_inserter", "~> 0.1"
+gem "fast_inserter", "~> 2.0"
```
jparr, line 92: I checked the changelog and this seems fine. it seems to have test coverage as well :)

app/controllers/smarty_streets_controller.rb
```diff
@@ -9,10 +9,13 @@ def street_address
...
-  def suggest
-    render json: get_suggestions_for_address(params), status: 200
-  end
+  # This endpoint appears unused (TheNotary 1/16)
+  # Lets remove it and see if anything breaks! (jparr 4/23)
+  # todo: remove if not used
+
+  # def suggest
+  #   render json: get_suggestions_for_address(params), status: 200
+  # end
```
jparr, line 18: action already had a comment that it was unused from 2016 so I figured why not do this? Up to you if you think this is a good idea though.

esoterik (approved):

Issue #913, opened 2024-04-08
2024 Upgrade TODOs, nice-to-haves, etc


esoterik:
This is intended to be a meta issue to capture any other additional issues or possible improvements that come up while upgrading, just so nothing gets lost. 

# TODOs

- [ ] Upgrade rubocop

# Bugs

# Nice-to-haves

- [ ] Remove database cleaner per [this comment](https://github.com/EFForg/action-center-platform/pull/910#issue-2228297157)

jparr:
I have two questions remaining about removing paperclip.

1. Partners have an attached image :logo. However the controller does not allow this here. https://github.com/EFForg/action-center-platform/blob/320e9ba37ce0d4aa05383cb986ce212b0aa6f033/app/controllers/admin/partners_controller.rb#L57

so its not currently working. Should I fix this or just remove the logo from the model?

2. TweetTarget automatically attatches a targets profile picture using the api. Does EFF have a twitter dev api account? I can migrate this but if you arent using this feature it would be better to just remove it.

jparr:
3. Another issue I am seeing are the database migrations. this will fail after removal of the gem: https://github.com/EFForg/action-center-platform/blob/2024-upgrades-main/db/migrate/20140504080222_add_attachment_featured_image_to_action_pages.rb

looks like you can add back the helper or rewrite the migration to do the same work. Some of the columns used by paperclip arent used by carrierwave by default.

jparr:
Ahoy has depreciation warnings in the rails 7 branch. We should upgrade it.

jparr:
Will need to migrate from secrets to credentials for rails 7. I shouldnt be too bad since you mostly use env variables.

PR #914, opened 2024-04-09
Upgrade Sentry

jparr:
see: #913

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -1,5 +1,3 @@
-unless Rails.application.secrets.sentry_dsn.nil?
-  Raven.configure do |config|
-    config.dsn = Rails.application.secrets.sentry_dsn
-  end
+Sentry.init do |config|
+  config.dsn = Rails.application.secrets.sentry_dsn
 end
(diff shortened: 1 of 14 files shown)


config/initializers/sentry.rb
```diff
@@ -1,5 +1,3 @@
-unless Rails.application.secrets.sentry_dsn.nil?
-  Raven.configure do |config|
-    config.dsn = Rails.application.secrets.sentry_dsn
-  end
+Sentry.init do |config|
+  config.dsn = Rails.application.secrets.sentry_dsn
```
jparr, line 2: Need to init in order not to error in tests. It won't send events if sentry_dsn is empty.

alexymik:
A minor heads up that our internal Sentry instance is Sentry 9.1.2 which doesn't support the new `sentry-ruby` gem.

jparr:
> Sentry 9.1.2



> A minor heads up that our internal Sentry instance is Sentry 9.1.2 which doesn't support the new `sentry-ruby` gem.

I'll mark this as a draft so it doesnt get merged by accident. @alexymik do you happen to know if upgrading the sentry-raven gem would be compatible? I havent found a compatibility list yet.

PR #919, opened 2024-04-18
Misc modernization

jparr:
ref: #913 

Draft until other PRs are merged - then rebase


jparr:
I wonder if we shouldnt just go straight to rails 7 over here: https://github.com/EFForg/action-center-platform/pull/920
doing the rails 6 upgrade in interim piecewise commit is useful but if we arent QAing it separately it could be better to just do that one.

PR #920, opened 2024-04-19
Upgrade to rails 7

jparr:
ref: #913

After everything else in 913 is done :)


jparr:
maybe squash a bunch of the rails 6 upgrade work into fewer commits.

PR #938, opened 2025-04-08
[WIP] Github actions

hartsick:
Decided to break this out into a PR against the gh-workflow because I'm making a lot of changes.


.github/workflows/rails_ci.yml
```diff
@@ -5,37 +5,43 @@
 # This workflow will install a prebuilt Ruby version, install dependencies, and
 # run tests and linters.
 name: "Ruby on Rails CI"
-on:
-  push:
-    branches: [ $default-branch, 2024-upgrades-main ]
-  pull_request:
-    branches: [ $default-branch, 2024-upgrades-main ]
+on: push
```
hartsick, line 8: this is intended to be temporary, to allow me to test this out. though I think running on every push would be useful - I was finding that pushes to an opened PR weren't triggering a re-run

.github/workflows/rails_ci.yml
```diff
@@ -5,37 +5,43 @@
...
 jobs:
   test:
     runs-on: ubuntu-latest
     env:
       RAILS_ENV: test
-      DATABASE_URL: "postgres://rails:password@localhost:5432/rails_test"
-      aws_access_key_id: test
-      aws_secret_access_key: test
+
+      db_host: localhost
```
hartsick, line 19: I'd like to convert these to be more canonical rails (eg DATABASE_PASSWORD) but for now I want to minimize changes and it's what we already have in secrets

.github/workflows/rails_ci.yml
```diff
@@ -5,37 +5,43 @@
...
+      db_password: password
     services:
       postgres:
         image: postgres:latest
         ports:
           - "5432:5432"
         env:
-          POSTGRES_DB: rails_test
-          POSTGRES_USER: rails
+          POSTGRES_DB: actioncenter_test # must be same as in database.yml for test.database
```
hartsick, line 29: this is the default database that the container creates when spun up. I was having issues with it being something other than what we need/create in the tests, which I'm still confused by—shouldn't be needed. but this fixes it for now.

Dockerfile
```diff
@@ -23,25 +23,25 @@ RUN apt-get update && \
 COPY package.json package-lock.json ./
 RUN npm install
 
-ADD Gemfile* ./
+COPY Gemfile* ./
```
hartsick, line 26: changed this to COPY both for security reasons and to improve caching (was slowing down my development when I had to do some container rebuilds). should have no noticeable impact besides container build speed

config/database.yml
```diff
@@ -1,8 +1,7 @@
 common: &common
-  adapter: <%= Rails.application.secrets.db_adapter %>
+  adapter: postgresql
```
hartsick, line 2: I'd like to move toward hardcoding things we don't need to easily modify—adapter was an easy one (since we need it in test). encoding might be, as well, but I left it alone because we don't actually appear to be setting it in deployed env. didn't want to touch it for now.

config/database.yml
```diff
@@ -12,16 +11,15 @@ common: &common
 
 development:
   <<: *common
-
-test: &test
-  <<: *common
-  database: <%= "#{Rails.application.secrets.db_database}-test" %>
+  database: <%= Rails.application.secrets.db_database || "actioncenter_development" %>
```
hartsick, line 14: I'd like for this to be actioncenter_development to differentiate from the prod database, but wanted to not break peoples' dev environments so kept previous behavior as default.

config/secrets.yml
```diff
@@ -60,19 +59,14 @@ development: *default
 
 test:
   <<: *default
-  db_adapter: <%= ENV["test_db_adapter"] %>
```
hartsick, line 63: since we use the same database container across environments, we no longer need the test_x environment variables. can just use the defaults from earlier in yml

config/secrets.yml
```diff
@@ -60,19 +59,14 @@ development: *default
...
-  db_password: <%= ENV["test_db_password"] %>
-  db_host: <%= ENV["TESTDB_PORT_5432_TCP_ADDR"] || ENV["test_db_host"] %>
-  db_port: <%= ENV["test_db_port"] %>
-  db_sslmode: <%= ENV["test_db_sslmode"] %>
-  db_sslrootcert: <%= ENV["test_db_sslrootcert"] %>
   congress_forms_url: "https://congressforms.eff.org"
   congress_forms_debug_key: "shh"
   fastly_api_key:
   sentry_dsn: ''
+  amazon_region: noop
```
hartsick, line 66: hardcode these for test env always, rather than setting in the test container

docker/entrypoint.sh
```diff
@@ -5,10 +5,6 @@ if [ "$DB_AUTO_MIGRATE" == "true" ]; then
   bin/rake db:migrate
 fi
 
-if [ "$TEST_DB_AUTO_MIGRATE" == "true" ]; then
```
hartsick, line 8: we're just loading schema now for tests, so shouldn't need to worry about test db migrations

.env.test
hartsick, line 1: I think this file was unused, and was misleading as a result

config/database.yml
hartsick: ok great. I'll do that! had based actioncenter off current db in .env.example, but if we don't care about that anyway then I'll just clear it entirely

.github/workflows/rails_ci.yml
hartsick: yeah good idea, will do!

PR #960, opened 2025-07-31
Put embeds behind a feature flag for now

esoterik:


creative38studio-stack (changes requested):
Fixes #958 (https://github.com/user-attachments/files/21817584/Mozilla-RK%40fonghayden239-1.pdf)Fixes #959

PR #974, opened 2026-03-18
Delete public user accounts

hartsick:
But keep the admins. 

`.delete_all` [intentionally skips activemodel callbacks](https://api.rubyonrails.org/v8.1.2/classes/ActiveRecord/Relation.html#method-i-delete_all)

I tested this locally on small set of test data. Would appreciate a lookover to make sure:
1.  there aren't any other accounts we should exclude from deletion
2. confirm we _shouldn't_ run callbacks on deletion (I skipped them because it felt too risky/destructive)
3. any other gotchas I'm missing

Task can be ran as `bin/rake users:delete_nonadmins`

### DIFF
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -48,4 +48,16 @@ namespace :users do
       puts "Granting admin status to #{email} failed."
     end
   end
+
+  desc "Delete non-admins"
+  task delete_nonadmins: :environment do |_t, args|
+    delete_count = User.where(admin: false).count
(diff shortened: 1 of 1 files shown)


esoterik (approved):
The analytics page in the admin still loads fine after running this and shows data, so I think this is good to merge! The deletion took < 30s on my local db, which is a dump from october 2023 or 2024, so hopefully shouldn't take too long on prod.

hartsick:
thank you!!

